PR #2, opened 2018-04-04
Export a User configuration entity

deviantintegral:
Since a site may have multiple mpx users, each associated with one or more mpx accounts, it makes sense to have "user" as it's own configuration object. Also, it means "accounts" can be totally exported with CMI, without worrying about accidentally exporting an mpx password to git.

For this PR, I've placed the menu item right under `Media`. However, once we add a second configuration entity, I will move it all into a config page for all mpx settings.

configuration drush site-install 2018-04-04 13-35-01

user example com drush site-install 2018-04-04 13-34-19

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
-# Media MPX for Drupal 8
+# Media mpx for Drupal 8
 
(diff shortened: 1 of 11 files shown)


deviantintegral:
Replaced by #3.

PR #3, opened 2018-04-04
Support creating MPX user and account config entities

deviantintegral:
These are fairly standard config entities, however I generated users with Drupal Console, and Accounts with Drush just so I could learn the differences.

Here is the mpx settings overview:

mpx settings drush site-install 2018-04-04 14-30-49

The user configuration entity:

user example com drush site-install 2018-04-04 13-34-19

The account configuration entity:

add a mpx account drush site-install 2018-04-04 14-31-41

The next PR will add account discovery and additional configuration to accounts.


juanolalla (approved):
Cool work!

PR #4, opened 2018-04-05
Add service definitions

deviantintegral:
Depends on #3 .

This PR adds service definitions and factory classes to simplify working with the mpx API. For example, lets say a developer wants to load all mpx accounts:

```php
<?php

/** @var \Drupal\media_mpx\DataObjectFactory $factory */
$factory = \Drupal::service('media_mpx.data_object_factory');
// $user is a User configuration entity.
$dof = $factory->forMpxObject($user, 'Access Data Service', 'Account', '1.0');

// This selects all accounts with no filters, we need to make the parameter optional in `select()`.
$fields = new \Lullabot\Mpx\DataService\ByFields();
$accounts = $dof->select($fields);
foreach ($accounts as $account) {
  // Access account properties like $account->getId(), $account->getTitle()...
}
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,6 +6,7 @@
   "license": "GPL-2.0+",
   "homepage": "https://www.drupal.org/project/media_mpx",
   "minimum-stability": "dev",
+  "prefer-stable": true,
   "support": {
     "issues": "https://www.drupal.org/project/issues/media_mpx",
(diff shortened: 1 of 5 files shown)


vordude (approved):
Looks good. Thanks for walking me through it!

PR #5, opened 2018-04-05
Initial media entity for media objects

deviantintegral:
Depends on #4 .

This PR adds support for configuring an mpx media type for media - the core object in mpx we get video data from.

To configure:

1. Create an mpx user and account under `/admin/config/media/mpx`.
1. Create a new media type under `/admin/structure/media` and select 'mpx Media' as the type. *Set a label first* as there is a core bug causing a 503 on the `#ajax` callback if the machine name isn't set (I need to find / create that bug).

Most complex fields are not supported yet - just primitive types and dates.

edit test drush site-install 2018-04-05 15-13-00

To test field mapping:

1. Add a datetime field to the media type you created.
1. Back at the main configuration form, set `The date that this content becomes available for playback.` to the field you created.
1. Add a new media item. In the `mpx Media` field, paste a URL such as `https://data.media.theplatform.com/media/data/Media/12345` with one valid for your configured mpx account.
1. On save, the date/time field should be populated.

Open questions:
- [ ] Should I remove the Account configuration entity, and merge it directly into the media type configuration? I feel like it helps make the UI simpler, but now that I think of it I wonder if it's always going to be a one-to-one relationship between accounts and media types.

### DIFF
--- a/src/Form/AccountForm.php
+++ b/src/Form/AccountForm.php
@@ -4,6 +4,7 @@ namespace Drupal\media_mpx\Form;
 
 use Drupal\Core\Entity\EntityForm;
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Url;
 
 /**
  * Mpx Account form.
@@ -35,6 +36,26 @@ class AccountForm extends EntityForm {
         'exists' => '\Drupal\media_mpx\Entity\MpxAccount::load',
       ],
       '#disabled' => !$this->entity->isNew(),
+    ];
+
+    $users = array_map(function ($entity) {
+      /** @var \Drupal\media_mpx\Entity\UserInterface $entity */
+      return $entity->label();
+    }, $this->entityTypeManager->getStorage('media_mpx_user')->loadMultiple());
+
+    if (empty($users)) {
+      $url = Url::fromRoute('entity.media_mpx_user.add_form')->toString() . '?destination=' . \Drupal::service('path.current')->getPath();
+      $this->messenger()->addError($this->t('<a href="@add-user">Create at least one mpx user</a> before creating accounts.', [
+        '@add-user' => $url,
+      ]));
+      return [];
+    }
+
+    $form['user'] = [
+      '#type' => 'select',
+      '#title' => $this->t('mpx user'),
+      '#description' => $this->t('Select an mpx user to see what accounts are available.'),
+      '#options' => $users,
     ];
 
     $form['status'] = [
--- a/src/Form/AccountForm.php
+++ b/src/Form/AccountForm.php
@@ -62,7 +62,7 @@ class AccountForm extends EntityForm {
 
     $form = $this->idLabelForm($form);
 
-    $users = $this->loadMpxUsers();
+    $users = $this->loadMpxUsers($form_state);
     if (empty($users)) {
       $url = Url::fromRoute('entity.media_mpx_user.add_form')->toString() . '?destination=' . \Drupal::service('path.current')->getPath();
       $this->messenger()->addError($this->t('<a href="@add-user">Create at least one mpx user</a> before creating accounts.', [
@@ -173,12 +173,6 @@ class AccountForm extends EntityForm {
       ],
     ];
 
-    // Set the currently selected user on the initial load.
-    if (!$form_state->hasValue('user')) {
-      reset($users);
-      $form_state->setValue('user', key($users));
-    }
-
     $this->fetchAccounts($form, $form_state);
     return $form;
   }
@@ -216,14 +210,24 @@ class AccountForm extends EntityForm {
   /**
    * Return all configured mpx user names, keyed by their config entity ID.
    *
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The current state of the form.
+   *
    * @return array
    *   An array of user names, keyed by the config entity ID.
    */
-  private function loadMpxUsers(): array {
+  private function loadMpxUsers(FormStateInterface $form_state): array {
     $users = array_map(function ($entity) {
       /** @var \Drupal\media_mpx\Entity\UserInterface $entity */
       return $entity->label();
     }, $this->entityTypeManager->getStorage('media_mpx_user')->loadMultiple());
+
+    // Set the currently selected user on the initial load.
+    if (!$form_state->hasValue('user')) {
+      reset($users);
+      $form_state->setValue('user', key($users));
+    }
+
     return $users;
   }
 
--- a/src/Form/AccountForm.php
+++ b/src/Form/AccountForm.php
@@ -6,7 +6,7 @@ use Drupal\Core\Entity\EntityForm;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Path\CurrentPathStack;
 use Drupal\Core\Url;
-use Drupal\media_mpx\DataObjectFactory;
+use Drupal\media_mpx\DataObjectFactoryCreator;
 use Lullabot\Mpx\DataService\ByFields;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
@@ -27,7 +27,7 @@ class AccountForm extends EntityForm {
   /**
    * The factory used to load mpx objects.
    *
-   * @var \Drupal\media_mpx\DataObjectFactory
+   * @var \Drupal\media_mpx\DataObjectFactoryCreator
    */
   protected $dataObjectFactory;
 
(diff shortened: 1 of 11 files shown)


juanolalla:
I get an error when I try to create an account at /admin/config/media/mpx/account/add:

Error: Class 'Lullabot\Mpx\Client' not found in Drupal\Component\DependencyInjection\Container->createService() at media_mpx/src/Form/AccountForm.php(53): Drupal\Component\DependencyInjection\Container->get('media_mpx.data_...')

I don't know why is that, the mpx library is in the Drupal vendor directory. I guess I'm missing something setting this up, can you help me?

deviantintegral:
Was this before or after setting up the merge plugin to pull in composer.json? If it was before, it wouldn't surprise me if the autoloader didn't automatically pick up the manually added library.

juanolalla:
I didn't use the merge plugin, I merged the vendor folder myself. I don't know exactly how to use the merge plugin in this workflow, to get this module in this branch with its dependencies. It would be nice if you could add that explanation to the ticket :smile:

deviantintegral:
1. Check out the module and the branch to your D8 module directory.
1. In Drupal's `composer.json`, add the path to the module's `composer.json` in the merge plugin settings, after the line that includes `core/composer.json`.
1. Run `composer update` in the same directory. This will update `composer.lock` and the vendor directory.

src/Form/AccountForm.php
```diff
@@ -4,38 +4,74 @@
...
-      '#type' => 'machine_name',
-      '#default_value' => $this->entity->id(),
-      '#machine_name' => [
-        'exists' => '\Drupal\media_mpx\Entity\MpxAccount::load',
-      ],
-      '#disabled' => !$this->entity->isNew(),
-    ];
+    $users = $this->loadMpxUsers($form_state);
+    if (empty($users)) {
+      $url = Url::fromRoute('entity.media_mpx_user.add_form')->toString() . '?destination=' . \Drupal::service('path.current')->getPath();
```
juampynr, line 67: :+1:

src/Form/AccountForm.php
```diff
@@ -46,6 +82,45 @@ public function form(array $form, FormStateInterface $form_state) {
...
+   * @return array
+   *   The accounts form item.
+   */
+  public function fetchAccounts(array &$form, FormStateInterface $form_state) : array {
+    $user_entity_id = $form_state->getValue('user');
+
+    /** @var \Drupal\media_mpx\Entity\UserInterface $user */
+    $user = $this->entityTypeManager->getStorage('media_mpx_user')->load($user_entity_id);
+
+    $accountFactory = $this->dataObjectFactory->forObjectType($user, 'Access Data Service', 'Account', '1.0');
```
juampynr, line 102: is this a factory method? If so, shouldn't this be from instead of for?

deviantintegral: Hm. Let's see.

> An account factory for the Access Data Service Account object, version 1.0

> An account factory from the Access Data Service Account object, version 1.0

I actually think I should rename `$this->dataObjectFactory` and possibly the class.. It's actually a _factory_ for a _factory_ (see the services.yml for how it works).

juampynr: Fine for me. Up to you.

deviantintegral: A few ideas here:

1. Rename `\Drupal\media_mpx\DataObjectFactory` to something like `\Drupal\media_mpx\DataObjectAdapter`, as it adapts Drupal's classes into the PHP library (though it's not a strict implementation of the adapter pattern).
1. Change `\Drupal\media_mpx\DataObjectFactory` to extend `\Lullabot\Mpx\DataService\DataObjectFactory`, just overriding the constructor method.
1. Use the decorator pattern, where we implement `\Drupal\media_mpx\DataObjectFactory` with a new `\Lullabot\Mpx\DataService\DataObjectFactoryInterface`.

What do you think @juampynr ?

juampynr:
> Rename \Drupal\media_mpx\DataObjectFactory to something like \Drupal\media_mpx\DataObjectAdapter, as it adapts Drupal's classes into the PHP library (though it's not a strict implementation of the adapter pattern).

Agree, let's not use adapter as it could lead to missunderstandings.

> Change \Drupal\media_mpx\DataObjectFactory to extend \Lullabot\Mpx\DataService\DataObjectFactory, just overriding the constructor method.

I like this one! Seems straightforward.

> Use the decorator pattern, where we implement \Drupal\media_mpx\DataObjectFactory with a new \Lullabot\Mpx\DataService\DataObjectFactoryInterface.

Not sure about this one, I would have to see a pull request to really ponder if it makes the dev experience clearer or not.

juampynr:
@deviantintegral got this error when I tried to create an account:

image

Looks like I am missing some dependencies. Debugging.

juampynr:
I had to run `highwire/drupal-psr-16` to fix the following error:

image

Which came from:
image

I had to run `composer require symfony/cache` to fix the following error, which cam from the same line as the previous error:

image

After fixing the above, I get this error when I open the form to add an account:

image

juampynr:
@deviantintegral do I need a working mpx account to test this pull request?

deviantintegral:
@juampynr odd, all of those dependencies are in the [composer.json on 8.x-1.x](https://github.com/Lullabot/media_mpx/blob/8.x-1.x/composer.json). Did you run composer update to get them?

Yes, you do need a working mpx account. I will send you a link to them.

deviantintegral:
>> Change \Drupal\media_mpx\DataObjectFactory to extend \Lullabot\Mpx\DataService\DataObjectFactory, just overriding the constructor method.

> I like this one! Seems straightforward.

I just realized that really makes the constructor awkward, as it has to combine the constructor parameters and the `forObjectType` parameters into one method, which is 6 parameters. And, you can't easily use the service container, as the `forObjectType` parameters are needed at runtime. It's the same problem with decorating the class too.

I think that leaves me with just coming up with a better name. "Data object" comes from mpx's documentation, so we probably should include that somehow.

* DataObjectFactoryGenerator
* DataObjectFactoryCreator
* DataObjectFactoryManager
* ...?

juampynr:
Managed to get the account form, but seeing these warnings:

image

juampynr:
As per the question:

```
 Should I remove the Account configuration entity, and merge it directly into the media type configuration? I feel like it helps make the UI simpler, but now that I think of it I wonder if it's always going to be a one-to-one relationship between accounts and media types.
```
I would merge it. If we ever find an scenario where we need to extract it, then we will be in a better place to make it in a way that it is intuitive.

juampynr:
Got this when I tried to create a media entity after configuring the date field mapping:

image

juampynr:
@deviantintegral can you try performing the steps that you suggested? Perhaps it makes sense that, before mapping happens, the entity has the date field as NULL. See this debugging line:

image

Here is the field mapping:

image

And here is the date field:

image

juampynr:
```
DataObjectFactoryGenerator
DataObjectFactoryCreator
DataObjectFactoryManager
```

No preference. You know the domain well so go for it and pick one.

deviantintegral:
I filed #17 for merging the account entity in.

You actually ran into something I filed over at https://github.com/Lullabot/mpx-php/issues/95. I have code in one of the other branches that catches that TypeError, let me cherry-pick here.

deviantintegral:
@juampynr this should fix the type errors for you.

juampynr:
After updating my local branch, it works! However, if I don't fill out a date when creating the media (because mapping should take care of it) I:

1. See this error for a second before the page refreshes and I see the created media page:
image

2. The date field is not populated:
image

deviantintegral:
Oh! Regarding the error handling, I realized I fixed that Friday and didn't backport it to this branch. This should let mpx errors properly bubble up now.

https://github.com/Lullabot/media_mpx/pull/5/commits/ea21d11d95e66c716a30ba5020755ef3b70557a4

juampynr:
Right, now I see the following error. Shouldn't this be handled as a form validation error?

image

juampynr:
Using https://data.media.theplatform.com/media/data/Media/6560426 works. I am merging this as it already gives a lot of functionality. See my suggestion above for handling the media not found exception in case you want to create a new pull request for that one.

deviantintegral:
Thanks! The problem I'm having with error handling is that there's no way in the source plugin to say "this is a form error". Instead, you're supposed to return NULL, which means at best you log the error but leave the data empty. I'm going to file an issue for it as it will take some research.

juampynr:
So there is no way to at least throw an exception and let it bubble up until it catches form validation?

deviantintegral:
The exception is being thrown, but it's the catching that's tricky.

PR #8, opened 2018-04-18
WIP: Support saving the complete media object

### LINKED ISSUE #7: Support storing Media subobjects in the key / value store
mpx Media objects are very complex, and don’t just contain scalars. For example, the `content` field:

- Is an array of `MediaFile` objects
  - Which itself contains an array of `releases`
    - And each `release` contains a public ID or `pid` which we want to store and load directly.

## Implementation plan
* Save MediaFiles as serialized classes in a key-value store.
* ~~At the least, store them in a separate table instead of filling up the default key / value table.~~ I've decided against this for now. The default table stores "collection" as well as an ID, and I think we can split this out later fairly easily if needed.
* Ensure related objects are deleted on entity delete.
* Consider using [entity_keyvalue](https://www.drupal.org/project/entity_keyvalue)
* Also store the Media objects themselves as a value and provide a `getMpxMedia()` method as an escape hatch for calling code in case they want the fully-typed Media object instead of going through mapped data.
* Consider wrapping the k/v store in a cache backend, if it seems like performance warrants it.

API users would call `getMetadata($media, 'content');` for example to return the array of MediaFiles, which would be retrieved from the key / value store.

## Alternatives we decided against
- Serialize into a text column in the entity table like the D7 module - too complex and fragile
- Entities for Everything (a MediaFile entity, a Release entity, etc) - Tons of work to store data we likely don't need.
- [json field](https://www.drupal.org/project/jsonb) or a JSON column in SQL - little need seen so far for querying on subobject properties, and there are database portability issues with json fields. Also, you have to manually add indexes for columns.

deviantintegral:
Fixes #7 and depends on #5 .

This PR uses the entity_keyvalue module to store the mpx Media object in the key / value store. See #7 for details of how we got here.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     "symfony/property-info": "^3.4"
   },
   "require-dev": {
-    "cweagans/composer-patches": "dev-master",
+    "cweagans/composer-patches": "^1.6.4",
     "behat/mink-selenium2-driver": "1.3.x-dev",
     "behat/mink-extension": "v2.2",
     "drupal/coder": "^8.2",
--- a/composer.json
+++ b/composer.json
@@ -12,6 +12,7 @@
     "source": "http://cgit.drupalcode.org/media_mpx"
   },
   "require": {
+    "drupal/entity_keyvalue": "^1.0",
     "symfony/cache": "^3.4",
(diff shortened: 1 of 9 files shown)


deviantintegral:
Closed in favour of #12 .

PR #12, opened 2018-04-20
Drush mpx-import command

### LINKED ISSUE #7: Support storing Media subobjects in the key / value store

### LINKED ISSUE #10: Add a batch ingest drush command
We need a drush command to import existing mpx media.

* The command should take a Media type ID to determine what to import for.
* After completion, the command should tell the user to start importing through notifications instead.
* We need a way to determine if a previously imported item has been deleted.

deviantintegral:
Depends on #8  and resolves #10 and resolves #7 .

This PR adds initial support for importing mpx media items. I ended up removing `entity_keyvalue`. It only works once an entity has been saved (so you have an entity id) and that was leading to some super hacky workarounds. Instead, we now store media items using the id provided by mpx, instead of an entity id.

Performance is subpar with this this PR. There's no http async or batch requests - I would like to address those in followups.

To test, create an mpx Media type, and run:

```
$ vendor/bin/drush mpxi mpx_video # mpx_video is the media type ID to import for.

Importing mpx_video media
=========================

    0 [░░░░░░░░░░░░░░░░░░░░░░░░░░░░]
```

You should then see media entities in the media listing.


src/Commands/MediaMpxCommands.php
```diff
@@ -0,0 +1,211 @@
...
+use Drupal\media_mpx\DataObjectFactoryCreator;
+use Drupal\media_mpx\Plugin\media\Source\Media;
+use Drush\Commands\DrushCommands;
+use Lullabot\Mpx\DataService\Access\Account;
+use Lullabot\Mpx\DataService\ByFields;
+use Lullabot\Mpx\DataService\Media\Media as MpxMedia;
+use Lullabot\Mpx\DataService\ObjectListIterator;
+
+/**
+ * A Drush commandfile.
```
juampynr, line 17: I guess that this was auto-generated.

src/Commands/MediaMpxCommands.php
```diff
@@ -0,0 +1,211 @@
...
+    $this->keyValueFactory = $keyValueFactory;
+  }
+
+  /**
+   * Import mpx objects for a media type.
+   *
+   * @param string $media_type_id
+   *   The media type ID to import for.
+   *
+   * @usage media_mpx-import mpx_video
```
juampynr, line 72: Whoa! I did not know about `@usage`.

src/Commands/MediaMpxCommands.php
```diff
@@ -0,0 +1,211 @@
...
+      // Save the mpx Media item so it's available in getMetadata() in the
+      // source plugin.
+      $this->keyValueFactory->get('media_mpx_media')->set($mpx_media->getId(), $mpx_media);
+
+      foreach ($results as $media) {
+        $media->save();
+        $reset_ids[] = $media->id();
+      }
+
+      $this->entityTypeManager->getStorage('media')->resetCache($reset_ids);
```
juampynr, line 107: Why do you need to clear cache at this time? Are you updating results in the same loop?

src/Commands/MediaMpxCommands.php
```diff
@@ -0,0 +1,211 @@
...
+   *   The media type object to load the source plugin for.
+   *
+   * @return \Drupal\media_mpx\Plugin\media\Source\Media
+   *   The source plugin.
+   */
+  private function loadMediaSource(MediaTypeInterface $media_type): Media {
+    /** @var \Drupal\media_mpx\Plugin\media\Source\Media $media_source */
+    $media_source = $media_type->getSource();
+    if (!($media_source instanceof Media)) {
+      throw new \RuntimeException(dt('@type is not configured as a mpx Media source.', ['@type' => $media_type->id()]));
```
juampynr, line 152: should this be `as an mpx` ?

juampynr:
Static review looks fine and CircleCI is happy. Testing manually.

juampynr:
Getting this:

image

Do I need to be under a VPN? Here are my account settings:

deviantintegral:
Ah yeah, I've seen that occasionally too. Usually just rerunning drush will get the results in time.

However, I'd worked around that when working on the notification listener by disabling Drupal's default timeout entirely. I've pulled that fix back into this branch.

juampynr:
Works now! However, see the status messages:

image

Can you fix that @deviantintegral? Then go ahead and merge this pull request.

PR #14, opened 2018-04-23
WIP: Importing mpx players

deviantintegral:
Depends on #13. This PR adds support for Player media entities, and makes our import and listen code reusable for any type of supported mpx object.

PR #13, opened 2018-04-20
WIP: Drush command to listen for mpx notifications

deviantintegral:
Depends on #12 and resolves #9.

This adds a drush command that populates a queue with notifications to process. Normally I wouldn't have bothered with a separate queue, but you _have_ to reload the object from mpx to get it's full data, so that adds enough of a performance hit to make it worth it.

As is though, performance isn't great, and it's on our side in the PHP code. It's ~450ms to import each media item, which can totally be improved on. However, I'd like to save performance work for a new pull request.

To pull in notifications, run `vendor/bin/drush mpxl mpx_video`. Then, run `vendor/bin/drush queue-run media_mpx_notification` to process them. I'll put up gifs and a screenshot when this is ready for review.

PR #18, opened 2018-04-24
Import mpx thumbnails

### LINKED ISSUE #9: Add a notification drush command
We need a drush command to listen to mpx notifications and act on them.

deviantintegral:
After looking at the total diff, it's still less than 1000 LOC so I think it's better to do as one PR to wrap this up.

This PR pins us to the branch at https://github.com/Lullabot/mpx-php/pull/102 as we need the interface it defines.

# Drush command to listen for notifications

Resolves #9 and replaces #13.

This adds a drush command that populates a queue with notifications to process. Normally I wouldn't have bothered with a separate queue, but you have to reload the object from mpx to get it's full data, so that adds enough of a performance hit to make it worth it.

As is though, performance isn't great, and it's on our side in the PHP code. It's ~450ms to import each media item, which can totally be improved on. However, I'd like to save performance work for a new pull request.

To pull in notifications, run vendor/bin/drush mpxl mpx_video. Then, run vendor/bin/drush queue-run media_mpx_notification to process them. I'll put up gifs and a screenshot when this is ready for review.

# Importing mpx players

Replaces #14.

Adds support for Player media entities, and makes our import and listen code reusable for any type of supported mpx object.

# Importing mpx thumbnails

Implements the media thumbnail APIs so thumbnails are downloaded to Drupal's file system and available for image styles and admin listings. Importing items with `drush mpxi` should pull in the thumbnails too.

media drush site-install 2018-04-24 15-21-18


composer.json
```diff
@@ -16,7 +16,7 @@
   "require": {
     "symfony/cache": "^3.4",
     "highwire/drupal-psr-16": "^1.0",
-    "lullabot/mpx-php": "dev-master",
+    "lullabot/mpx-php": "dev-fix-account",
```
juampynr, line 19: This can be changed later as the branch has been merged.

src/Commands/NotificationQueuer.php
```diff
@@ -0,0 +1,266 @@
...
+   * Return the current notification ID, or -1 if one is not set.
+   *
+   * @param \Drupal\media\MediaSourceInterface $media_source
+   *   The media source the notification ID is for.
+   *
+   * @return int
+   *   The notification ID.
+   */
+  protected function getNotificationId(MediaSourceInterface $media_source): int {
+    // @todo should this really be state?
```
juampynr, line 131: I think so. We don't want this to be exported to configuration, nor to exist in other sites than the current one, right?

src/Plugin/QueueWorker/NotificationQueueWorker.php
```diff
@@ -0,0 +1,111 @@
...
+      $container->get('media_mpx.data_object_factory_creator'),
+      $container->get('media_mpx.data_object_importer')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function processItem($data) {
+    /* @var $data \Drupal\media_mpx\Notification[] */
```
juampynr, line 86: Just curious, why not adding this to the docblock?

src/Plugin/QueueWorker/NotificationQueueWorker.php
```diff
@@ -0,0 +1,111 @@
...
+      $media_source = $this->importer::loadMediaSource($notification->getMediaType());
+      $factory = $this->dataObjectFactoryCreator->fromMediaSource($media_source);
+      $promises[] = $factory->load($mpx_media->getId());
+      $media_types[] = $notification->getMediaType();
+    }
+
+    // Process each request concurrently.
+    // @todo Handle individual request rejections by requeuing them to the
+    // bottom of the queue.
+    each_limit($promises, 10, function ($mpx_media, $index) use ($media_types) {
```
juampynr, line 106: Whoa! This one is new to me.

https://api.drupal.org/api/drupal/vendor%21guzzlehttp%21promises%21src%21functions.php/function/each_limit/8.5.x

juampynr:
Code review looks fine. Testing manually.

juampynr:
Probably unrelated: got this warning right after adding the mpx account:

image

juampynr:
Worked! Here is what I did:

1. Configured the MPX media and MPX player media types.
2. Imported MPX media:

image

3. Listened a few times for notifications:

image

4. Verified that media was imported along with thumbnails:

image

5. Imported players:

image

6. Verified that players were imported:

image

src/Commands/NotificationQueuer.php
deviantintegral: I think it's going to depend on if there are separate "stage" accounts in mpx or not. Either way, the worst case is we reprocess some data, and our notification system always uses the current state of the data, not whatever the (stale) notification might say.

src/Plugin/QueueWorker/NotificationQueueWorker.php
deviantintegral: Last I did that, years ago, the Drupal code standards threw errors.

deviantintegral:
Good catch on the warning - that is totally a mistake from when I aliased the class. I must have left the "check strings" option enabled.

PR #22, opened 2018-04-27
Player formatter

deviantintegral:
This PR:

* Cleans up some code debt in notifications with some refactoring of the NotificationQueuer into a second NotificationListener class.
* Supports adding a player formatter for videos.
* Stores the account public IDs on the account entities to save some complexity in rendering players.

To test:

* Re-save any mpx account config entities to ensure they have a public ID saved.
* In the field display options, set the mpx URL to the `mpx player` formatter.
* View a video!

Note that players may not be 100% functional if they require additional javascript that is not provided by the player iframe. For this PR, I consider "success" to be "we have an iframe with a rendered player inside of it".


src/Plugin/Field/FieldFormatter/PlayerFormatter.php
```diff
@@ -0,0 +1,194 @@
...
+
+    $factory = $this->dataObjectFactoryCreator->forObjectType($source_plugin->getAccount()->getUserEntity(), 'Player Data Service', 'Player', '1.6');
+    $player = $factory->load($this->getSetting('player'))->wait();
+
+    foreach ($items as $delta => $item) {
+      /** @var \Lullabot\Mpx\DataService\Media\Media $mpx_media */
+      $mpx_media = $source_plugin->getMpxObject($entity);
+      $url = new Url($source_plugin->getAccount(), $player, $mpx_media);
+
+      // What cache contexts or tags do we set?
```
juampynr: Can't we define custom tags for mpx clips? Then, if a clip is updated, we would invalidate them.

juampynr:
I could see and configure the view mode for the mpx media entity. Looks like the following:

image

However, when I try to view an mpx media entity I see this error:

image

I see that it's coming from here, shall I debug anything else?

image

src/Plugin/Field/FieldFormatter/PlayerFormatter.php
deviantintegral: I'm not sure if we need to define anything custom - since they are media entities, the tags should be available. I'm just not aware if the FAPI adds them automatically (probably does, just want to be sure!). I will label this as `@todo` so it shows up.

deviantintegral:
Ah good catch. I changed that over at https://github.com/Lullabot/mpx-php/commit/8c32b9007ac310f824f38b9024fc4090b77a9687.

I think the library is stable enough to start tagging - I will tag a `0.1.0` release and add that to this PR.

juampynr:
Still getting the same error even though I updated locally both this module's branch and the `mpx-php` library.

Now having the following setup:

image

Here is the error:

image

Do you want me to do some debugging?

deviantintegral:
No, thanks for looking again. I think it's time I write a Behat test or three.

juampynr:
Two weeks ago I gave JavaScriptTestBase another go now that you can use WebDriver. They work fine! for testing a module I think that it is a great choice.

See the [change record for instructions](https://www.drupal.org/node/2941464) plus [representative_module](https://www.drupal.org/project/representative_image) for a few test examples.

deviantintegral:
@juampynr This PR has grown as I pulled in mock responses, as well as a new submodule to mock requests for JS (or Behat) tests. See https://github.com/Lullabot/media_mpx/pull/22/commits/649b06cb13766c033852300f3adfc95b48d44983 which failed before fixing https://github.com/Lullabot/media_mpx/pull/22/commits/fcd68c2b4cb397a78fd06d8b25ccdef1111180e7 (which was what you reported above).

test_modules/media_mpx_test/src/MediaMpxTestServiceProvider.php
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Service provider to replace the mpx client with a mock client.
+ */
+class MediaMpxTestServiceProvider extends ServiceProviderBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function alter(ContainerBuilder $container) {
```
juampynr, line 17: :+1:

juampynr:
No more errors! I don't see a player but I do see an iframe:

image

Let's get this in so we can move forward.

PR #26, opened 2018-04-30
Cache class metadata

deviantintegral:
This improves performance by using cached data for the Symfony deserializer. Instead of having to parse classes every time to determine what properties they have, this lets them persist across requests. Since the classes are code, any changes to them imply a cache clear on deployment.


juampynr:
Last time I click that button without editing the message. This time for real.

PR #29, opened 2018-05-01
Media type migrations

deviantintegral:
This migrates:

* mpx accounts
* mpx users
* Creates media types for each account
* Migrates media associated with each media type

Run them all with `drush mim --group=media_theplatform_mpx`. By default, this will migrate from data in a 'migrate' database configured in settings.php, just like any other migration.


juampynr:
@deviantintegral, this is still listed as WIP. Are you still working on it?

juampynr:
@deviantintegral can you fix conflicts in this branch?

juampynr:
Oh wow, I like code climate:

image

I think that I am going to add that to https://github.com/Lullabot/drupal8ci/

deviantintegral:
FYI @juampynr , it's not conflicts, but I've turned on the setting that requires feature branches to be updated with master. You can use the "Update branch" button or merge locally if you'd rather.

deviantintegral:
@juampynr I refactored to simplify the return statements, this PR should be good to review now.

PR #31, opened 2018-05-03
Add a command to warm the kv storage

deviantintegral:
~~Depends on https://github.com/Lullabot/mpx-php/pull/107~~

I wanted a way to locally cache mpx data without having to run a full migration. This may also be useful for new sites in letting them load mpx data without worrying about the 0.5-1s "first request" performance hit.


vordude (approved):
Works for me I suppose.  No actual use of the warmKeyValue command (yet) I take it?

juampynr:
conflicts

deviantintegral:
I've used it locally to make my local faster. For example, I left it running in the background while working on the actual migration code, and then testing the migration was faster because we had the data already stored. I will probably create a SQL dump of the data at some point as a nice way to bootstrap environments.

PR #40, opened 2018-05-18
Custom field discovery

deviantintegral:
See the changes in the README for details for custom module implementations. The test module should provide a good example of how this works. The discovery class was mostly copied from AnnotatedClassDiscovery in core, but significantly refactored because that class is painfully complex.


vordude (approved):
Untested, and somewhat understood, but I can't hold it up. :+1:

PR #41, opened 2018-05-22
Validate credentials and handle mpx connection errors

### LINKED ISSUE #16: Catch credential exceptions when triggered from the UI
These should be caught, logged, and printed as a message.

deviantintegral:
This PR resolves #16 by:

* Checks the username and password when saving the mpx user form.
* Adds error handling catching HTTP errors to the User, Account, and Player Formatter Settings forms.
* Handles errors when rendering players, logging the exception and continuing to the next item in the list.

### DIFF
--- a/src/Form/AccountForm.php
+++ b/src/Form/AccountForm.php
@@ -7,7 +7,11 @@ use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Path\CurrentPathStack;
 use Drupal\Core\Url;
 use Drupal\media_mpx\DataObjectFactoryCreator;
+use Drupal\media_mpx\MpxLogger;
+use GuzzleHttp\Exception\TransferException;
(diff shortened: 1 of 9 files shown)


src/Form/AccountForm.php
```diff
@@ -233,19 +250,72 @@ private function loadMpxUsers(FormStateInterface $form_state): array {
    *     - The account options.
    *     - The account public IDs.
    */
-  private function accountOptions(FormStateInterface $form_state): array {
-    $user_entity_id = $form_state->getValue('user');
+  protected function accountOptions(FormStateInterface $form_state): array {
+    try {
+      try {
```
juampynr, line 255: Interesting: I have never seen a try within a try.

deviantintegral: Yeah, I try to avoid it, but it's the only way to handle catching an exception and then deciding you need to re-throw it for a more general handler.

Issue #23, opened 2018-04-29
Set a time limit for tokens


deviantintegral:
We should limit them to something reasonable, like 5 minutes or an hour, just so if they are ever disclosed the risk is minimal.

PR #42, opened 2018-05-23
Set a 5 minute time limit for tokens #23

deviantintegral:
Depends on https://github.com/Lullabot/mpx-php/pull/125

- [x] Update composer.json


juanolalla (approved):
:+1:

Issue #21, opened 2018-04-27
Fix imported media being owned by Anonymous


deviantintegral:
Which leaves the question, who should it be owned by?

juampynr:
THE PEOPLE! :-D

PR #44, opened 2018-05-23
Fix imported media being owned by Anonymous #21

deviantintegral:
I took a look at a D7 site, and it has the same issue - imported content is owned by user 0. I think user 1 is a much more reasonable default. If needed, we can add a config setting later to set which user ID imported media gets saved as.


juampynr:
All is good. Merging.

PR #46, opened 2018-05-25
Replace the key/value store with a cache

### LINKED ISSUE #43: Fix class serialization errors when custom fields change
On average, I'm seeing around 500ms response times from mpx to load a single media object. That adds up, especially when importing objects locally. I wanted a way to test importing without always hitting mpx, so `\Drupal\media_mpx\DataObjectImporter::setKeyValue` was added to store deserialized mpx objects in the key-value store.

I'd explicitly left this out of the cache system as I wanted to be able to rebuild caches _without_ clearing my locally stored mpx data.

However, this breaks in one critical case - during integration with custom fields. For example, let's say we create a `lullabot_mpx` module with a custom field implementation and turn it on. When we load the object, it will still have the stub `MissingCustomFieldsClass` set on the data. Or, lets say mpx doesn't change their object schema, but the mpx library adds a new method or internal field. We won't have that data, and FATALs result.

I see a few ways we could solve this:

1. Stop using the k/v store, and fall back to a straight cache implementation.
1. Instead of storing deserialized objects in k/v, store the literal mpx response data.
1. Do both, and store mpx response data in the cache (instead of serialized objects).
1. Create a "developer only" cache bin backend that is the opposite of the `\Drupal\Core\Cache\NullBackend` class - instead of never caching, it overrides `\Drupal\Core\Cache\CacheBackendInterface::deleteAll` to not delete anything. This would probably live as it's own developer module, and be enabled in `development.services.yml` and `settings.php`.

deviantintegral:
Fixes #43 .

This PR changes how we store mpx responses. Instead of storing serialized classes, we store the literal http responses from mpx. That gives us much more flexibility, such as being able to reload objects into different classes (such as adding a new custom field implementation) without having to query mpx again.

~~Depends on https://www.drupal.org/project/guzzle_cache/issues/2945390 (though I'm patching this in) and https://github.com/Lullabot/mpx-php/pull/126~~

### DIFF
--- a/media_mpx.services.yml
+++ b/media_mpx.services.yml
@@ -20,7 +20,7 @@ services:
 
   media_mpx.data_object_importer:
     class: Drupal\media_mpx\DataObjectImporter
-    arguments: ['@keyvalue', '@entity_type.manager']
+    arguments: ['@entity_type.manager', '@cache.media_mpx_http']
 
   media_mpx.notification_listener:
     class: Drupal\media_mpx\NotificationListener
@@ -65,7 +65,14 @@ services:
 
   media_mpx.client_factory:
     class: Drupal\media_mpx\ClientFactory
-    arguments: ['@media_mpx.http_handler_stack', '@http_client_factory']
+    arguments: ['@media_mpx.http_handler_stack', '@http_client_factory', '@cache.media_mpx_http']
+
+  cache.media_mpx_http:
+    class: Drupal\Core\Cache\CacheBackendInterface
+    tags:
+      - { name: cache.bin }
+    factory: cache_factory:get
+    arguments: [media_mpx_http]
 
   # These services are "internal" and generally not needed by custom code.
   media_mpx.session_lock:
--- a/src/ClientFactory.php
+++ b/src/ClientFactory.php
@@ -3,8 +3,12 @@
 namespace Drupal\media_mpx;
 
 use Drupal\Component\Utility\NestedArray;
+use Drupal\Core\Cache\CacheBackendInterface;
 use Drupal\Core\Http\ClientFactory as HttpClientFactory;
+use Drupal\guzzle_cache\DrupalGuzzleCache;
 use GuzzleHttp\HandlerStack;
+use Kevinrob\GuzzleCache\CacheMiddleware;
+use Kevinrob\GuzzleCache\Strategy\Delegate\DelegatingCacheStrategy;
 use Lullabot\Mpx\Client;
 use Lullabot\Mpx\Middleware;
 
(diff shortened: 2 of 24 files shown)


media_mpx.services.yml
```diff
@@ -65,7 +65,14 @@ services:
 
   media_mpx.client_factory:
     class: Drupal\media_mpx\ClientFactory
-    arguments: ['@media_mpx.http_handler_stack', '@http_client_factory']
+    arguments: ['@media_mpx.http_handler_stack', '@http_client_factory', '@cache.media_mpx_http']
+
+  cache.media_mpx_http:
```
deviantintegral, line 70: This splits the responses into a separate cache bin, and core handles creating any database tables automatically. Neat!

src/ClientFactory.php
```diff
@@ -34,14 +38,33 @@ class ClientFactory {
...
     $this->handlerStack = $handlerStack;
 
     // This must be in the constructor, otherwise this will be added every time
     // a new mpx client is created during a page request.
     $this->handlerStack->push(Middleware::mpxErrors(), 'mpx_errors');
+
+    $cache = new DrupalGuzzleCache($cacheBackend);
+
+    $strategy = new DelegatingCacheStrategy();
+    $strategy->registerRequestMatcher(new MpxObjectRequestMatcher(), new MpxCacheStrategy($cache, 3600 * 24 * 30));
```
deviantintegral, line 55: By default, we cache objects for 30 days. Sites should get fresh data by listening to notifications.

src/Commands/MpxImporter.php
```diff
@@ -129,13 +127,16 @@ public function warmKeyValue(string $media_type_id) {
...
-    each_limit($list->yieldLists(), 4, function (ObjectList $list) use ($media_type, $importer, &$count) {
-      foreach ($list as $item) {
-        $importer->setKeyValue($item, $media_type);
+    $service_info = $media_source->getPluginDefinition()['media_mpx'];
+
+    each_limit($list->yieldLists(), 4, function (ObjectList $list) use ($media_type, $service_info) {
+      foreach ($list as $index => $item) {
+        // For each item, we need to inject it into the response cache as if
+        // a single request for the item was made.
+        $this->dataObjectImporter->cacheItem($item, $service_info);
```
deviantintegral, line 139: This is our hack to fill the cache without re-loading each object again from mpx.

juampynr (approved):
Looks good @deviantintegral. Do you want me to do any manual testing? If so, what should I do?

deviantintegral:
You could try running the `mpxw` drush command, but I just added a commit that shows the caches working by removing duplicate http responses for the same request.

PR #48, opened 2018-05-28
Fix migration key

deviantintegral:
This fixes hardcoding the database key, as well as a missing migrate dependency.

### DIFF
--- a/migrations/d7_media_theplatform_mpx_source_field_instance.yml
+++ b/migrations/d7_media_theplatform_mpx_source_field_instance.yml
@@ -3,6 +3,7 @@ label: Source field instance for mpx objects
 migration_group: media_theplatform_mpx
 migration_tags:
   - 'Drupal 7'
+  - 'Configuration'
 migration_dependencies:
   required:
     - d7_media_theplatform_mpx_source_field_storage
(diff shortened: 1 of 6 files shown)


migrations/d7_media_theplatform_mpx_source_field_instance.yml
```diff
@@ -1,6 +1,11 @@
 id: d7_media_theplatform_mpx_source_field_instance
 label: Source field instance for mpx objects
 migration_group: media_theplatform_mpx
+migration_tags:
+  - 'Drupal 7'
```
juampynr, line 5: Can you add the Configuration tag? TugBoat runs these first.

migrations/d7_media_theplatform_mpx_source_field_storage.yml
```diff
@@ -1,6 +1,10 @@
 id: d7_media_theplatform_mpx_source_field_storage
 label: Source field for mpx objects
 migration_group: media_theplatform_mpx
+# This migration cannot be attached to the Drupal 7 tag as it uses the Empty
+# plugin, which doesn't define the required source_module key.
+#migration_tags:
+#  - 'Drupal 7'
```
juampynr, line 7: Is there a ticket for that? Can I help?

deviantintegral: I think this is a core issue. Empty literally means "no source" but core assumes all migrations from Drupal 7 have a source. I'm starting to think that perhaps we're overusing this tag in our migrations.

migrations/d7_media_theplatform_mpx_source_field_instance.yml
deviantintegral: Done!

PR #50, opened 2018-05-31
Dispatch an import event

### LINKED ISSUE #34: Add an event for imports to alter a ByFields class
Custom code may want to filter out mpx assets. The best way to do this is to filter them out mpx-side, saving both ends time.

Before calling `select`, we should dispatch an event with the current `ByFields` object. Then, custom code can add or remove conditions as it sees fit.

Unfortunately, notifications can't be filtered in this way. Instead, I think we will need something on the Drupal side to filter out objects.

deviantintegral:
Resolves #34.

See the README for what this event is for and how it can be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Note that this event should _not_ be used for mapping mpx data to Drupal
 fields. Instead, use the mapping functionality in the media type configuration.
 
 See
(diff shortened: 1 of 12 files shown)


juampynr:
Looks great!

PR #51, opened 2018-06-01
Fix typo at field instance migration

juampynr:


deviantintegral:
I totally fixed this upstream and forgot to fix it in our forked configuration 🤦‍♂️ 

Thanks!

PR #52, opened 2018-06-01
Fix drush bootstrap

deviantintegral:
This PR:

* Adds an update hook to ensure the new guzzle_cache dependency is enabled.
* Handles class not found FATALs in Drush's bootstrap if the module hasn't been enabled yet. IMO this is a bug or design issue in Drush; it basically makes it impossible to use services from other contrib modules, as they may add new dependencies that break drush completely.


media_mpx.install
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * @file
+ * Installation hooks for media_mpx.
+ */
+
+/**
+ * Install the new guzzle_cache dependency.
+ */
+function media_mpx_update_8101() {
+  \Drupal::service('module_installer')->install(['guzzle_cache']);
```
juampynr, line 12: :+1:

juanolalla (approved):
Yes, it fixes it, great work!!! Merging, thanks!

PR #53, opened 2018-06-06
Support altering imports with Q Queries

deviantintegral:
This will need to be updated once https://github.com/Lullabot/mpx-php/pull/129 is merged and a new tag is out.

### DIFF
--- a/src/Commands/MpxImporter.php
+++ b/src/Commands/MpxImporter.php
@@ -9,9 +9,9 @@ use Drupal\media_mpx\DataObjectImporter;
 use Drupal\media_mpx\Event\ImportSelectEvent;
 use Drush\Commands\DrushCommands;
 use function GuzzleHttp\Promise\each_limit;
-use Lullabot\Mpx\DataService\ByFields;
 use Lullabot\Mpx\DataService\ObjectList;
 use Lullabot\Mpx\DataService\ObjectListIterator;
+use Lullabot\Mpx\DataService\ObjectListQuery;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 
 /**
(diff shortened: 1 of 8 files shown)


juanolalla:
Oops, unit-kernel tests are failing

deviantintegral:
Fixed @juanolalla !

src/Commands/MpxImporter.php
```diff
@@ -157,10 +157,10 @@ public function warmCache(string $media_type_id) {
   private function selectAll(MediaTypeInterface $media_type): ObjectListIterator {
     $media_source = DataObjectImporter::loadMediaSource($media_type);
     $factory = $this->dataObjectFactoryCreator->fromMediaSource($media_source);
-    $fields = new ByFields();
-    $event = new ImportSelectEvent($fields, $media_source);
+    $query = new ObjectListQuery();
+    $event = new ImportSelectEvent($query, $media_source);
     $this->eventDispatcher->dispatch(ImportSelectEvent::IMPORT_SELECT, $event);
-    $results = $factory->select($fields, $media_source->getAccount());
+    $results = $factory->select($query, $media_source->getAccount());
```
juanolalla, line 163: I see you're changing here the argument passed to $factory->select, but I can't see you had changed that in the select() definition.

juanolalla (approved):
Good work, just a question to understand it better.

src/Commands/MpxImporter.php
deviantintegral: Ah yes, `select()` is in the mpx-php library. This PR is addressing the API changes between the 0.3.0 and 0.4.0 releases. See https://github.com/Lullabot/mpx-php/pull/129/files

PR #54, opened 2018-06-08
Fix TypeError when invalid credentials are used on the AccountForm

deviantintegral:
If a password is wrong, we return NULL instead of the array as documented.


juampynr:
Looks good. Merging.

PR #55, opened 2018-06-08
Support mapping MediaFile properties

deviantintegral:
This PR adds specific support for mapping MediaFile properties.

Some properties, notably duration, are not on the Media object, but on each individual MediaFile. Each Media object has a `content` field containing an array of media files, which could be thumbnails, videos, or audio.

Since we can't know which index to use (easily), we find the first video and use that. This is what the Drupal 7 code does and works well enough. If we find a need to expose that logic to custom code, I think we can dispatch an event or similar. But, hopefully this is enough!


juampynr:
Looks good and everything passes. Merging.

PR #61, opened 2018-07-23
Update PlayerFormatter.php

juampynr:
FormatterBase.php in Drupal 8.6 uses `protected` as the access modifier for the `$messenger` property, while this class uses private. This causes the following PHP error:

```
[Mon Jul 23 12:43:16.434148 2018] [php7:notice] [pid 1896] [client 127.0.0.1:38106] PHP Fatal error:  Access level to Drupal\\media_mpx\\Plugin\\Field\\FieldFormatter\\PlayerFormatter::$messenger must be protected (as in class Drupal\\Core\\Field\\FormatterBase) or weaker in /var/www/somesite/d8/web/modules/contrib/media_mpx/src/Plugin/Field/FieldFormatter/PlayerFormatter.php on line 35, referer: http://somesite.local/admin/structure/types/manage/blog/display-layout/teaser
```

Here is a screenshot:

image

### DIFF
--- a/src/Plugin/Field/FieldFormatter/PlayerFormatter.php
+++ b/src/Plugin/Field/FieldFormatter/PlayerFormatter.php
@@ -60,7 +60,7 @@ class PlayerFormatter extends FormatterBase implements ContainerFactoryPluginInt
    *
    * @var \Drupal\Core\Messenger\MessengerInterface
    */
-  private $messenger;
+  protected $messenger;
 
   /**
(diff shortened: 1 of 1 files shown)


deviantintegral:
Thanks! Glad this works for 8.5 as well.

deviantintegral:
Looks like two different commits were pushed with the same content, and I merged the other one. This is fixed now.

PR #64, opened 2018-09-17
Update migration files

juampynr:
* Uses a post_row_save to create the source field for media bundles.
* Uses the file_managed table as the base for migrating media.
* Creates derivatives with mappings for each bundle field.


juampynr:
Moving it to https://www.drupal.org/project/media_mpx

PR #65, opened 2018-09-21
Migrate adjustments

juampynr:
Makes some adjustments to migrate users, accounts, media bundles,
and content migrated into a site.

* Uses a post_row_save to create the source field for media bundles.
* Uses the file_managed table as the base for migrating media.
* Creates derivatives with mappings for each bundle field.


src/Plugin/migrate/MpxItemDeriver.php
```diff
@@ -0,0 +1,257 @@
...
+      $container->get('plugin.manager.migrate.cckfield'),
+      $container->get('plugin.manager.migrate.field'),
+      $container->get('plugin.manager.migration')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getDerivativeDefinitions($base_plugin_definition) {
```
m4olivei, line 100: @juampynr I broke this method down into parts to help with readability and to appease the Code Climate job. Could you double check that I didn't break things?

deviantintegral:
This looks fine to me, merging!

juampynr:
Thanks @m4olivei!

PR #69, opened 2018-10-10
Issue #68 Make private members of \Drupal\media_mpx\Plugin\Field\FieldFormatter\PlayerFormatter protected

### LINKED ISSUE #68: \Drupal\media_mpx\Plugin\Field\FieldFormatter\PlayerFormatter makes extending hard
I'd like to extend `\Drupal\media_mpx\Plugin\Field\FieldFormatter\PlayerFormatter` to create a new  field formatter based on it for rendering a playlist rather than just the single video. The member visibility makes that difficult in a few places, namely the injected services members, but it's also conceivable that a class extending it might also want to customize the `fetchPlayerOptions` and `renderIframes` methods, all of which is currently private.

m4olivei:
Resolves #68


deviantintegral (approved):
Makes sense to me.

Issue #63, opened 2018-07-30
Fix only storing one notification ID for all media bundles


deviantintegral:
The notification ID stored in the state table should be per media type, not global for all media types.

https://github.com/Lullabot/media_mpx/blob/8.x-1.x/src/NotificationListener.php is where the code is that needs to be fixed.

geekygnr:
@deviantintegral I think I have figured out what I need to do.

Would you like the notification id to include the media source plugin id as well as the media type id or is the media type id on its own sufficient?

deviantintegral:
I think the media type is fine, since a media type's source can't change after it's been created.

PR #70, opened 2018-10-11
Fix only storing one notification ID for all media bundles #63

geekygnr:
notification id's now have the bundle id's embedded in them.

Solution:
Changed the function parameters to take in the media type id instead of the media source and use it to generate the notification id.

Fixes:
https://github.com/Lullabot/media_mpx/issues/63


deviantintegral (approved):
Looks great! Will merge once the master merge build passes.

Issue #59, opened 2018-07-11
Track object's updated time when processing the notification queue


deviantintegral:
When listening for notifications, we already filter out duplicate notifications. However, it turns out we still end up with many, many notifications for the same object over multiple notification responses from mpx (in contrast to duplicates within the same single notification). This makes processing the notification queue significantly slower than it should be.

To work around this, we should leverage our cached mpx data and check the last updated time before we process a notification. Something like:

1. When saving a notification, store the timestamp of the update.
1. Load the mpx object.
1. Check it's updated date. If it's older than the notification timestamp, go ahead and process the notification. Otherwise, discard the notification.

geekygnr:
For testing this issue from @deviantintegral 

> You can use the video titled `Captain Lee Rosbach Calls the RHONY Boat Ride from Hell Drama "Overblown"` , it's full ID is `http://data.media.theplatform.com/media/data/Media/1342208067763`

geekygnr:
Pull request made so I am going to close this.

PR #72, opened 2018-10-18
Track object's updated time when processing the notification queue

geekygnr:
Resolves: https://github.com/Lullabot/media_mpx/issues/59

Solved this by comparing the changed time of the entities to the date on the notification.


src/Commands/NotificationQueuer.php
```diff
@@ -82,6 +85,11 @@ public function listen($media_type_id) {
     $this->io()->note(dt('Waiting for a notification from mpx after notification ID @id...', ['@id' => $notification_id]));
     $notifications = $this->listener->listen($media_source, $notification_id);
     $notifications = $this->filterDuplicateNotifications($notifications);
+    $notifications = $this->filterByDate($notifications, $media_type_id);
+
+    if (empty($notifications)) {
```
deviantintegral, line 90: 👍

src/Commands/NotificationQueuer.php
```diff
@@ -92,6 +100,57 @@ public function listen($media_type_id) {
...
 
+  /**
+   * Removes notifications that are older than the entities that reference them.
+   *
+   * @param \Lullabot\Mpx\DataService\Notification[] $notifications
+   *   An array of notifications.
+   * @param string $media_type_id
+   *   The media type being imported.
+   *
+   * @return array
```
deviantintegral: This returns `\Lullabot\Mpx\DataService\Notification[]`? If we document that then IDEs can do autocomplete on the array contents automatically.

src/Commands/NotificationQueuer.php
```diff
@@ -92,6 +100,57 @@ public function listen($media_type_id) {
...
+   *   The filtered array.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\PluginException
```
deviantintegral: No need to document `@throws` unless we actually throw it ourselves.

src/Commands/NotificationQueuer.php
```diff
@@ -92,6 +100,57 @@ public function listen($media_type_id) {
...
+
+    /** @var \Drupal\Media\MediaStorage $media_storage */
+    $media_storage = \Drupal::entityManager()->getStorage('media');
+    /** @var \Drupal\Media\Entity\MediaType $media_type */
+    $media_type = MediaType::load($media_type_id);
+    $source = $media_type->getSource();
+    $source_field = $source->getSourceFieldDefinition($media_type)->getName();
+
+    $notifications = array_filter($notifications, function (MpxNotification $notification) use ($media_storage, $source_field) {
+      $notificationDate = $notification->getEntry()->getUpdated()->format("U");
```
deviantintegral, line 126: Do we need to update `mpx-php` as well for the getUpdated method on ObjectInterface?

src/Commands/NotificationQueuer.php
```diff
@@ -92,6 +100,57 @@ public function listen($media_type_id) {
...
+      $notificationId = (string) $notification->getEntry()->getId();
+      $entities = $media_storage->loadByProperties([$source_field => $notificationId]);
+
+      // The updates must have to do with something new so keep them included.
+      if (empty($entities)) {
+        return TRUE;
+      }
+
+      /* If there exists an entity that hasn't been updated since the
```
deviantintegral, line 135: I'm surprised this passes phpcs, but normally the convention is `//` for inline coments. But if it passes, it passes :D

src/Commands/NotificationQueuer.php
```diff
@@ -92,6 +100,57 @@ public function listen($media_type_id) {
...
+      notification was updated keep the notification. */
+      foreach ($entities as $entity) {
+        /** @var \Drupal\Media\Entity\Media $entity */
+        if ($entity->getChangedTime() < $notificationDate) {
```
deviantintegral, line 140: This is fine, but lets add a new issue for handling the case where something inside of Drupal or an editor has manually saved the media entity, and link that here with `@todo`.

deviantintegral (commented):
Looks great! A few minor notes, this will be good to merge once those are done.

src/Commands/NotificationQueuer.php
geekygnr: Ya that would make sense. Looking at the the interface compared to all the implemented functions on the classes, it looks like there is a lot that could be added to the interface. Did you want my to make a ticket to have the interface updated with all the common and useful public functions? And if so, do you want that to be pre-beta work?

src/Commands/NotificationQueuer.php
geekygnr: https://github.com/Lullabot/media_mpx/issues/75

src/Commands/NotificationQueuer.php
deviantintegral: Yes, let's fix it. Otherwise we have the potential for FATALs down the line:

notificationqueuer php drupal8 2018-10-25 10-22-55

I would say certainly add updated, and do a quick look for anything else I've missed. It doesn't need to be very in-depth.

deviantintegral:
Let's get the missing method warning cleaned up before merging this, otherwise 👍

geekygnr:
> Let's get the missing method warning cleaned up before merging this, otherwise +1

https://github.com/Lullabot/mpx-php/pull/145

PR #74, opened 2018-10-24
Retry thumbnails when they fail to download

deviantintegral:
There's an upstream core issue at https://www.drupal.org/project/drupal/issues/1832818#comment-12826680. I'd like to get feedback there before finalizing this.

Testing:

- [ ] Import a video from mpx, ensuring the bundle is set to queue thumbnail downloads.
- [ ] Make sure you don't run the queue, and then break your mpx calls by changing the mpx password to something invalid.
- [ ] Run cron. An error will be logged and the queue item should be recreated once in the media_entity_thumbnail queue.

TODO:

- [x] Document that `drush queue-run` isn't affected by this patch.
- [x] File an upstream issue to refactor out core's processQueues method so it can be used by drush?
  * Done: https://www.drupal.org/project/drupal/issues/3008997

### DIFF
--- a/src/Plugin/QueueWorker/ThumbnailDownloader.php
+++ b/src/Plugin/QueueWorker/ThumbnailDownloader.php
@@ -40,7 +40,7 @@ class ThumbnailDownloader extends CoreThumbnailDownloader {
       // If core has been patched to support recreating a single queue item,
       // use that.
       // @see https://www.drupal.org/project/drupal/issues/1832818
-      if (!class_exists('\Drupal\Core\Queue\PostponeItemException')) {
+      if (class_exists('\Drupal\Core\Queue\PostponeItemException')) {
         throw new PostponeItemException($message, $e->getCode(), $e);
       }
 
--- a/src/Plugin/QueueWorker/ThumbnailDownloader.php
+++ b/src/Plugin/QueueWorker/ThumbnailDownloader.php
@@ -2,7 +2,7 @@
 
 namespace Drupal\media_mpx\Plugin\QueueWorker;
 
-use Drupal\Core\Queue\RecreateException;
+use Drupal\Core\Queue\PostponeItemException;
 use Drupal\Core\Queue\SuspendQueueException;
 use Drupal\media\Entity\Media;
 use Drupal\media\Plugin\QueueWorker\ThumbnailDownloader as CoreThumbnailDownloader;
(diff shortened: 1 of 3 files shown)


deviantintegral:
@geekygnr mind giving this a review / test when you're next working on this? Thanks!

src/Plugin/QueueWorker/ThumbnailDownloader.php
```diff
@@ -0,0 +1,84 @@
...
+        $media->id(),
+        $media->getSource()->getSourceFieldValue($media),
+      ]);
+
+      // If core has been patched to support recreating a single queue item,
+      // use that.
+      // @see https://www.drupal.org/project/drupal/issues/1832818
+      if (!class_exists('\Drupal\Core\Queue\PostponeItemException')) {
+        throw new PostponeItemException($message, $e->getCode(), $e);
+      }
```
geekygnr, line 42: Am I reading this right? If the class does not exist throw that class?

src/Plugin/QueueWorker/ThumbnailDownloader.php
```diff
@@ -0,0 +1,84 @@
...
+   *
+   * @param array $data
+   *   The queue data.
+   *
+   * @return bool
+   *   TRUE if data references an mpx media entity, FALSE otherwise.
+   */
+  private function isMpxMediaItem(array $data): bool {
+    /** @var \Drupal\media\Entity\Media $media */
+    $media = $this->getMedia($data);
```
geekygnr: Instead of calling getMedia here why not get rid of the getMedia function and load the media object above line 24. That way you are only making one call to load this media item and spare adding a function to the execution stack. This function would then accept a Media object or null instead of the data array.

geekygnr (commented):
One spot where either I am reading the code wrong or you are negating something when you shouldn't and on efficiency suggestion that you can take or leave really.

src/Plugin/QueueWorker/ThumbnailDownloader.php
deviantintegral: Whoops. Good catch.

src/Plugin/QueueWorker/ThumbnailDownloader.php
deviantintegral: 👍

Issue #30, opened 2018-05-01
Decide how to handle unavailable notification IDs


deviantintegral:
Should this just throw an error, or should we transparently rerun notification with `-1`?

geekygnr:
PR made

PR #76, opened 2018-10-25
If the notification id is old re-run the query with a notification id of -1

geekygnr:
Resolves: https://github.com/Lullabot/media_mpx/issues/30

### DIFF
--- a/src/NotificationListener.php
+++ b/src/NotificationListener.php
@@ -105,7 +105,7 @@ class NotificationListener {
           'The last notification ID %id for %account is older than 7 days and is too old to fetch notifications. The last notification ID has been reset to re-start ingestion of all videos.',
           [
             '%id' => $notification_id,
-            '%account' => $account->getPid(),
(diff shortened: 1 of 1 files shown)


src/NotificationListener.php
```diff
@@ -99,6 +100,17 @@ public function listen(MpxMediaSourceInterface $media_source, int $notification_
         return [];
       }
 
+      if ($e->getCode() == 404) {
+        $this->logger->warning(
+          'The last notification ID %id for %account is older than 7 days and is too old to fetch notifications. The last notification ID has been reset to re-start ingestion of all videos.',
+          [
+            '%id' => $notification_id,
+            '%account' => $account->getPid(),
```
deviantintegral: Since we have the media source, can we make this easier to trace? The public id is a random set of letters. Instead, it could be `$media_source->getAccount()->label()` which would be the label of the account config entity.

PR #77, opened 2018-10-26
Support availability entity access

deviantintegral:
This depends on https://github.com/Lullabot/mpx-php/pull/146 so that should be merged first, and then the composer.json updated.


m4olivei (commented):
Also looks great. Nice and clean.

Issue #24, opened 2018-04-29
Use the batch API for ingestion


deviantintegral:
While it may not be strictly needed, it would help protect us from any small memory leaks over a long ingestion.

deviantintegral:
I was really hoping to avoid this, but I've run an import for almost an hour and memory has gone from a start of ~120MB to almost 300MB, for ~23000 items. I'm sure we could eliminate some leaks, but there's enough code we don't control that we need to either batch or queue things.

I took a look at the D8 batch API and it's not improved much since D7. IOW, because it uses arrays for everything it's really easy for little mistakes to break things. While queues have a little more admin complexity, they also make it possible to run concurrently. Since I'll need to rework things, I am going towards the queue direction.

juampynr:
Makes sense to me, @deviantintegral.

geekygnr:
A PR has been made so I am closing this.

PR #81, opened 2018-10-27
Change the mpx importing to be handled by a queue #24

geekygnr:
This has been done to manage memory issues.
Resolves https://github.com/Lullabot/media_mpx/issues/24
This issue is dependent on https://github.com/Lullabot/mpx-php/pull/147

### DIFF
--- a/src/Commands/MpxImporter.php
+++ b/src/Commands/MpxImporter.php
@@ -115,9 +115,8 @@ class MpxImporter extends DrushCommands {
 
     $results = $this->select($media_type, $query);
 
-    // @todo Support fetching the total results via ObjectList.
     $this->io()->title(dt('Importing @type media', ['@type' => $media_type_id]));
-    $this->io()->progressStart();
+    $this->io()->progressStart($this->getTotalCount($media_type, $query));
 
     $queue = $this->queueFactory->get(self::MEDIA_MPX_IMPORT_QUEUE);
 
@@ -126,7 +125,7 @@ class MpxImporter extends DrushCommands {
       $queue->createItem(new MpxImportTask($mpx_media->getId(), $media_type_id));
 
       $this->io()->progressAdvance();
-      $this->logger()->info(dt('Imported @type @uri.', ['@type' => $media_type_id, '@uri' => $mpx_media->getId()]));
+      $this->logger()->info(dt('@type @uri has been queued to be imported.', ['@type' => $media_type_id, '@uri' => $mpx_media->getId()]));
     }
 
     $this->io()->progressFinish();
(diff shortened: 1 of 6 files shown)


geekygnr:
Not sure what to do about the code climate issue. I was following the same pattern as the notification importer which has similar function definitions.

deviantintegral:
Looks like there's two issues flagged:

One is forced on us by Drupal's code - we can't reduce the constructor even if we wanted to. Once you create an account on Code Climate and log in, you should have permission to mark these sorts of issues as "wontfix".

media_mpx - code climate 2018-10-29 13-24-03

The other one is completely in our own code, so I'm going to take a quick look and see if there's something obvious for it.

src/Commands/MpxImporter.php
```diff
@@ -150,14 +176,18 @@ public function warmCache(string $media_type_id) {
...
    * @param \Drupal\media\MediaTypeInterface $media_type
    *   The media type to load items for.
+   * @param \Lullabot\Mpx\DataService\ObjectListQuery $query
+   *   An optional query to limit what is returned.
    *
    * @return \Lullabot\Mpx\DataService\ObjectListIterator
    *   An iterator over all retrieved media items.
    */
-  private function selectAll(MediaTypeInterface $media_type): ObjectListIterator {
+  private function select(MediaTypeInterface $media_type, ObjectListQuery $query = NULL): ObjectListIterator {
```
deviantintegral, line 184: 👍

src/Plugin/QueueWorker/MpxImporterQueueWorker.php
```diff
@@ -0,0 +1,114 @@
...
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entityTypeManager
+   *   An entity type manager.
+   * @param \Drupal\media_mpx\DataObjectImporter $data_importer
+   *   The data importer.
+   * @param \Drupal\media_mpx\DataObjectFactoryCreator $dataObjectFactoryCreator
+   *   Data factory creator.
+   */
+  public function __construct(array $configuration, string $plugin_id, $plugin_definition, EntityTypeManagerInterface $entityTypeManager, DataObjectImporter $data_importer, DataObjectFactoryCreator $dataObjectFactoryCreator) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->entityTypeManager = $entityTypeManager;
```
deviantintegral, line 81: A small thing, while Drupal is OK with either camel case or underscores, it's a good practice not to mix and match them in the same function or class.

deviantintegral (commented):
Static review looks pretty good, I'm going to test this out now.

PR #84, opened 2018-10-30
Fix empty notifications after filtering never setting the last ID

deviantintegral:
If all notifications are filtered out by date, and we are left with an empty array of notifications, we never set the last notification ID. That means it's left at `-1`, and we process the same batch of notifications over and over again, never progressing.

### DIFF
--- a/src/Commands/NotificationQueuer.php
+++ b/src/Commands/NotificationQueuer.php
@@ -84,20 +84,29 @@ class NotificationQueuer extends DrushCommands {
     // of an mpx object in a row).
     $this->io()->note(dt('Waiting for a notification from mpx after notification ID @id...', ['@id' => $notification_id]));
     $notifications = $this->listener->listen($media_source, $notification_id);
+
+    // Keep track of the initial count of notifications so we can know if we
+    // filtered down to an empty array.
+    $initial_count = count($notifications);
+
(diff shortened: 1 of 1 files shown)


src/Commands/NotificationQueuer.php
```diff
@@ -84,20 +84,29 @@ public function listen($media_type_id) {
...
+    $last_notification = end($notifications);
+
     $notifications = $this->filterDuplicateNotifications($notifications);
     $notifications = $this->filterByDate($notifications, $media_type_id);
 
-    if (empty($notifications)) {
-      return;
+    if (empty($notifications) && $initial_count) {
+      $this->io()->note(dt('All notifications were skipped as newer data has already been imported.'));
+    }
```
geekygnr, line 101: Would it reduce the Complexity if we add a return here, took out the else and moved line 105 to before this if statement.

I know it is within the limit now but the past few tickets this is all I have been thinking about.

geekygnr (approved):
+1

src/Commands/NotificationQueuer.php
deviantintegral: I actually started with that locally, but then you run into trouble with calling `setNotificationId` that needs to happen in both cases.

PR #86, opened 2018-11-01
Issue #85 field formatter settings

### LINKED ISSUE #85: Add auto play and play all toggle to mpx player field formatter
The upstream `lullabot/mpx-php` library supports setting the `autoPlay` and `playAll` query string parameters on the player URL: https://github.com/Lullabot/mpx-php/blob/master/src/Service/Player/Url.php#L107,L129.

Can we reflect those as field formatter settings?

m4olivei:
Resolves #85.


src/Plugin/Field/FieldFormatter/PlayerFormatter.php
```diff
@@ -329,6 +348,8 @@ private function addMediaFileDetails(array &$element, MpxMedia $mpx_media) {
    */
   private function buildPlayer(Media $source_plugin, Player $player, MpxMedia $mpx_media) {
     $url = new Url($source_plugin->getAccount(), $player, $mpx_media);
+    $url->setAutoplay((bool) $this->getSetting('auto_play'));
```
deviantintegral: Why do we need the casting here? Only thing I can think of would be if we defaulted these settings to `NULL`, allowing us to not include the parameters at all if they aren't set. However, that's not going to be possible as soon as you save the field formatter, which brings me back to the casting bit.

deviantintegral (changes requested):
Looks like tests are failing, can you investigate?

src/Plugin/Field/FieldFormatter/PlayerFormatter.php
m4olivei: You're right, the casting isn't needed. I think this was a relic from before I introduced the config schema and these values were coming through as strings, eg. `'1'` and `'0'`.

deviantintegral (approved):
🎉

Issue #36, opened 2018-05-09
Add schema.org values to our default player template


deviantintegral:
Here is an example from Drupal 7:

```
<div class="video-clip" itemprop="video" itemscope itemtype="http://schema.org/VideoObject">
  <meta itemprop="name" content="<?php print $video_title; ?>" />
  <meta itemprop="description" content="<?php print $video_desc; ?>" />
  <meta itemprop="duration" content="<?php print $video_duration; ?>" />
  <meta itemprop="thumbnailUrl" content="<?php print $video_thumbnail; ?>" />
  <meta itemprop="embedURL" content="<?php print $video_src; ?>" />
  <meta itemprop="uploadDate" content="<?php print $video_available; ?>" />
  <meta itemprop="height" content="<?php print $video_height; ?>" />
  <meta itemprop="width" content="<?php print $video_width; ?>" />
  <?php print $iframe['#children']; ?>
</div>
```

geekygnr:
Pull request created

Issue #33, opened 2018-05-09
Add a class to iframes for behaviours to use


deviantintegral:
We need to support behaviours on iframes, so custom code can react to player events. To do that, we need appropriate classes on the iframes so they can be selected with behaviours.

We should also document in the README how to implement a behaviour and respond to a player event.

geekygnr:
@deviantintegral I put in a couple of initial classes based on a guess of the kind of identifiers that would be needed (See https://github.com/Lullabot/media_mpx/tree/iframe-classes).

My initial thought was:

```
'mpx-iframe',
'mpx-iframe-account--' . $source_plugin->getAccount()->id(),
```
The only other thing I could think of added was one with the pid of the media object so there is something identifying the specific object. I haven't done that yet since it sounds like there are better ways to identify things based on other discussions we have had.

One thing I noticed while trying to identify what you wanted in the README and if there was anything else I should do to support the use of behaviours I found this which suggests an HTML5 player should be possible. Looking at our implementation I found that the there is an option under the advanced settings tab. I haven't tested to avoid doing something I shouldn't but it looks like you can prioritize HTML over flash.
https://docs.theplatform.com/help/players-embed-players-using-iframe-elements

As for the behaviours, I am not 100% what to document but I found this. Is this the behaviour you would like documented to the README and if so are there any rules around disclosure I should be aware of? https://docs.theplatform.com/help/player-using-the-external-controller

deviantintegral:
I like the idea of per-account classes, though I'd probably use the machine name of the account config entity instead.

`iframe` in the class name seems redundant. How about just `mpx-player` and `mpx-player-account--<machine_name>`?

Yes, HTML5 players are not only possible but used in most cases. However, there is still an iframe that is used to host whatever player is being returned.

I was referring to Drupal behaviours - just so that developers can know we add those classes without having to go splunking through the code to find them.

geekygnr:
Pull request created

PR #87, opened 2018-11-02
Player template adjustments

geekygnr:
Resolves
- #33 
- #36 

Some processing on the the render array will need to happen to add some of the meta data that is specific to the site. Only meta data that doesn't change on a per site basis has been added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@
   - [2. Altering ingested mpx objects](#2-altering-ingested-mpx-objects)
 - [Custom field support](#custom-field-support)
 - [Migrating from Media: thePlatform mpx](#migrating-from-media-theplatform-mpx)
+- [Interacting with the Video Player](#interacting-with-the-video-player)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
@@ -196,3 +197,13 @@ machine name is `mpx_lullabot`, add the following to `settings.php`:
 
 $config['media_mpx.media_mpx_user.mpx_lullabot']['password'] = 'SECRET';
 ```
+## Interacting with the Video Player
+
+The video player for media objects will be rendered as an iframe which in
+turn utilizes a player from MPX. If this iframe needs to be interacted with
+through Drupal behaviours the following classes are added to the iframe.
+ - `mpx-player`
+ - `mpx-player-account--<Account Machine Name>`
+ 
+ Additional classes as well as other HTML attributes can be added to the iframe
+ by modifying the render array for the `media_mpx_iframe` template.
--- a/README.md
+++ b/README.md
@@ -200,10 +200,10 @@ $config['media_mpx.media_mpx_user.mpx_lullabot']['password'] = 'SECRET';
 ## Interacting with the Video Player
 
 The video player for media objects will be rendered as an iframe which in
-turn utilizes a player from MPX. If this iframe needs to be interacted with
+turn utilizes a player from mpx. If this iframe needs to be interacted with
 through Drupal behaviours the following classes are added to the iframe.
  - `mpx-player`
  - `mpx-player-account--<Account Machine Name>`
- 
- Additional classes as well as other HTML attributes can be added to the iframe
- by modifying the render array for the `media_mpx_iframe` template.
+
+Additional classes as well as other HTML attributes can be added to the iframe
+by modifying the render array for the `media_mpx_iframe` template.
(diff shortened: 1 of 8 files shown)


geekygnr:
I don't think it makes sense to do the refactoring it suggests. What do you think @deviantintegral?

README.md
```diff
@@ -196,3 +197,13 @@ machine name is `mpx_lullabot`, add the following to `settings.php`:
 
 $config['media_mpx.media_mpx_user.mpx_lullabot']['password'] = 'SECRET';
 ```
+## Interacting with the Video Player
```
deviantintegral, line 200: 👍

src/Element/MediaMpxIframeWrapper.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+namespace Drupal\media_mpx\Element;
+
+use Drupal\Core\Render\Element\RenderElement;
+use Drupal\Core\Template\Attribute;
+
+/**
+ * Providers an element design for wrapping media iframes.
```
deviantintegral: Typo in `Providers`

src/Element/MediaMpxIframeWrapper.php
```diff
@@ -0,0 +1,46 @@
...
+   *
+   * @param array $element
+   *   An element array before being processed.
+   *
+   * @return array
+   *   The processed and renderable element.
+   */
+  public static function preRenderInlineFrameEmbed(array $element) {
+    if (is_array($element['#attributes'])) {
+      $element['#attributes'] = new Attribute($element['#attributes']);
```
deviantintegral: Ooo interesting, I thought core handled `#attributes` for you automatically this way.

src/Plugin/Field/FieldFormatter/PlayerFormatter.php
```diff
@@ -244,23 +247,98 @@ protected function renderIframes(FieldItemListInterface $items, Player $player,
...
+      '#type' => 'media_mpx_iframe_wrapper',
+      '#attributes' => [
+        'class' => [
+          'mpx-iframe-wrapper',
+        ],
+      ],
+      '#meta' => [
+        'name' => $entity->label(),
+        'thumbnailUrl' => $thumbnail_url,
+        'uploadDate' => $mpx_media->getAvailableDate()->format(DATE_ISO8601),
```
deviantintegral, line 290: Yay, glad this works.

src/Plugin/Field/FieldFormatter/PlayerFormatter.php
```diff
@@ -244,23 +247,98 @@ protected function renderIframes(FieldItemListInterface $items, Player $player,
...
+      ],
+      '#content' => $this->buildPlayer($source_plugin, $player, $mpx_media),
+    ];
+    $this->addMediaFileDetails($element, $mpx_media);
+
+    return $element;
+  }
+
+  /**
+   * Adds the video specific details to schema.org meta data.
```
deviantintegral: `metadata` is usually one word.

I'm going to note that we pull from the first file.

tests/src/FunctionalJavascript/PlayerFormatterTest.php
```diff
@@ -113,7 +113,7 @@ public function testViewPlayerFormatter() {
     $this->getSession()->getPage()->fillField('Name', $this->randomString());
     $this->getSession()->getPage()->fillField('mpx Media', 'http://data.media.theplatform.com/media/data/Media/2602559');
     $this->click('#edit-submit');
-    $this->assertSession()->responseContains('<iframe src="https://player.theplatform.com/p/public-id/DemoPlayer/select/media/Zy1n9JQPy6hw"></iframe>');
+    $this->assertSession()->responseContains('<iframe class="mpx-player mpx-player-account--' . $account->id() . '" src="https://player.theplatform.com/p/public-id/DemoPlayer/select/media/Zy1n9JQPy6hw"></iframe>');
```
deviantintegral, line 116: You got this working, nice!

deviantintegral (commented):
This looks great. I'm going to do some final touchups so we can get this finished in our current sprint.

src/Element/MediaMpxIframeWrapper.php
deviantintegral: Hah, you got this from my code I wrote back in April, so clearly I ran into this already.

deviantintegral:
I think it's worth pulling a base class for the render elements, given that it's simple (5 minutes, not the hour code climate estimates) and that the only difference is the theme value.

PR #89, opened 2018-11-05
Support description, duration, and embedUrl values

deviantintegral:
I was working on implementing this on a site, and I discovered:

* While the site was mapping descriptions to Drupal fields, the edit forms are set so they aren't editable. In other words, the data is still read-only.
* In thinking through that, I changed my mind on how to handle attributes that could be overridden. Instead of not including them, we should include default reasonable values from mpx (like we do for other values). Sites can override those defaults with Drupal fields if they so choose. To support that, I've added entity and mpx_media variables to the theme hook.

### DIFF
--- a/media_mpx.module
+++ b/media_mpx.module
@@ -42,6 +42,8 @@ function media_mpx_theme() {
         'attributes' => [],
         'meta' => [],
         'content' => [],
+        'entity' => NULL,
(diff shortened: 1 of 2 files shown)


deviantintegral:
@geekygnr I'm merging, but feel free to leave comments.

PR #92, opened 2018-11-12
Issue #91 Make all members of \Drupal\media_mpx\Plugin\Field\FieldFormatter\PlayerFormatter protected instead of private

### LINKED ISSUE #91: \Drupal\media_mpx\Plugin\Field\FieldFormatter\PlayerFormatter makes extending hard redux
Coming up against the same issue as in #68. This time, specifically I'd like to see the `buildMeta` method be protected so that I can call it independently in a child class. It's conceivable a child class might like to pick out other pieces, so lets make all private members protected for extendability.

m4olivei:
Resolves #91.

### DIFF
--- a/src/PlayerMetadata.php
+++ b/src/PlayerMetadata.php
@@ -34,7 +34,6 @@ class PlayerMetadata {
    */
   protected $playerUrl;
 
-
   /**
    * Constructs a PlayerMetadata object.
    *
--- a/src/PlayerMetadata.php
+++ b/src/PlayerMetadata.php
@@ -1,0 +1,74 @@
+<?php
+
+namespace Drupal\media_mpx;
+
+use Drupal\media\Entity\Media as DrupalMedia;
+use Lullabot\Mpx\DataService\Media\Media as MpxMedia;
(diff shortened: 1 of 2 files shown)


deviantintegral:
A thought, would it make more sense to split generating the metadata out into a new class? It could have a `toArray` method and so on.

m4olivei:
> split generating the metadata out into a new class

@deviantintegral were you thinking something like this?

src/PlayerMetadata.php
```diff
@@ -0,0 +1,73 @@
...
+   * hook, which are the keys for scheme.org tags output with an MPX player.
+   *
+   * @return array
+   *   Metadata keys for schema.org tags.
+   */
+  public function toArray() {
+    $source_plugin = $this->drupalMedia->getSource();
+    return [
+      'name' => $this->drupalMedia->label(),
+      'thumbnailUrl' => file_create_url($source_plugin->getMetadata($this->drupalMedia, 'thumbnail_uri')),
```
deviantintegral, line 66: I really wish we could have an OOP API for this!

deviantintegral (approved):
Excellent!

PR #99, opened 2018-12-11
Implement a caching mechanism for the API

juampynr:
Thumbnail processing, field mapping, xmlsitemap access, and
many others will end up calling getMpxObject() several times
for the same entity.

This change suggests that, during the same request, we could
cache the result of the MPX API, reducing the number of
requests per media entity to 1.

### DIFF
--- a/media_mpx.services.yml
+++ b/media_mpx.services.yml
@@ -66,10 +66,25 @@ services:
     class: Drupal\media_mpx\ClientFactory
     arguments: ['@media_mpx.http_handler_stack', '@http_client_factory', '@cache.media_mpx_http']
 
+  # Defines an array or memory cache that supports a limit on the number of
+  # items.
+  cache.backend.media_mpx.array_cache_pool:
+    class: Drupal\media_mpx\ArrayCachePoolFactory
+    arguments: ['@datetime.time', '@cache_tags.invalidator.checksum', 100]
+
+  # Use the above array cache on top of the normal HTTP request cache. We use
+  # the ChainedFastBackend cache so any invalidations on the consistent cache
+  # are shared with all threads.
+  cache.backend.media_mpx.chainedfast_http:
+    class: Drupal\Core\Cache\ChainedFastBackendFactory
+    arguments: ['@settings', null, cache.backend.media_mpx.array_cache_pool]
+    calls:
+      - [setContainer, ['@service_container']]
+
   cache.media_mpx_http:
     class: Drupal\Core\Cache\CacheBackendInterface
     tags:
-      - { name: cache.bin }
+      - { name: cache.bin , default_backend: cache.backend.media_mpx.chainedfast_http}
     factory: cache_factory:get
     arguments: [media_mpx_http]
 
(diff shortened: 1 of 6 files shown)


juampynr:
Notice that I have taken the simplest approach possible just to prove my point. Just to give you an example of how this affects something like a Drupal to Drupal migration, migrating 100 items without this patch takes   300 seconds, while applying this patch reduces it to 60.

m4olivei:
Hum, my understanding was that there already is a caching layer in place as a Guzzle handler:

https://github.com/Lullabot/media_mpx/blob/8.x-1.x/src/ClientFactory.php#L54,L67

Does this mean that isn't effective?

juampynr:
This change reduced considerably migration times. I will get a screenshot in a few minutes.

juampynr:
This is weird. While I saw a massive improvement in performance locally, processing times at https://circleci.com/gh/nbc-bravo/d8/32584 are not much different from what they were before I will wait for this complete and tomorrow I will create a pull request without the patch and compare results to decide whether I should revert my change or not.

deviantintegral:
Yes, we have caching, but we don't have a static in-memory cache. I was a little worried about adding one because mpx objects have a decent amount of data in them. If we want to inject a memory cache, we could investigate using `\Drupal\Core\Cache\ChainedFastBackend` , `\Drupal\Core\Cache\MemoryCache\MemoryCache`, and `\Doctrine\Common\Cache\ChainCache.

m4olivei:
Selfishly, I'd love to see such a combination in practice, sounds neat.

m4olivei:
At a glance, it doesn't look like `\Doctrine\Common\Cache\ChainCache` will accept the Drupal cache objects, is there a way around that?

juampynr:
After hitting a timeout while migrating content, I restored the patch locally and performed a migration of 100 MPX entities while the xdebug profiler was on. Here is the command output where you can see the processing time and memory consumption:

image

And here is another run with the patch. Notice that, while it uses more memory (makes sense as it creates an array of MPX objects), it takes half the time to complete:

image

Looking at kCacheGrind graphs, here is one without the patch. Notice that while we are processing 100 elements, we are hitting the MPX API almost 300 times:

image

In contrast, here is a graph with the patch applied, where we hit the API a 100 times:

image

Unfortunately I cannot dig deeper into the caching layer to find whether there is a bug or not, but I will once I get a chance.

juampynr:
I just added a check to limit the cache array to clear itself once it contains a 100 elements. I had to do this as otherwise during a migration it would eat more and more memory. Once I did this change, I started seeing consistent processing times and memory usage:

image

deviantintegral:
> At a glance, it doesn't look like \Doctrine\Common\Cache\ChainCache will accept the Drupal cache objects, is there a way around that?

No, unfortunately Drupal core didn't use any upstream interfaces for D8 caches. I used https://github.com/Capgemini/drupal-doctrine-cache in D7 and it'd be trivial to do the same for D8. Adapter pattern FTW!

src/Plugin/media/Source/MediaSourceBase.php
```diff
@@ -201,11 +201,23 @@ public function getAccount(): Account {
...
+      $cache = [];
+    }
+
     $id = $media->get($this->configuration['source_field'])->getString();
     $factory = $this->dataObjectFactoryCreator->fromMediaSource($this);
 
-    return $factory->load(new Uri($id))->wait();
+    $cache[$media->id()] = $factory->load(new Uri($id))->wait();
 
+    return $cache[$media->id()];
```
deviantintegral: This is always going to return `TRUE`, because `id()` on a new entity returns null, which gets cast to an empty string. This leads to broken data on imports as all new media will have the same values as the first media item imported.

I'll work on fixing this.

deviantintegral (dismissed):

src/Plugin/media/Source/MediaSourceBase.php
juampynr: Whoa! Great catch @deviantintegral.

deviantintegral:
@m4olivei I've rewritten this branch to inject an array cache before each backend cache. Feel free to ping if you want a walkthrough.

deviantintegral:
One note is that this PR is ~10% slower than the prior implementation because it's using the consistent backend. If that's a problem later we can swap it to the inconsistent backend.

media_mpx.services.yml
```diff
@@ -66,10 +66,25 @@ services:
...
+  cache.backend.media_mpx.array_cache_pool:
+    class: Drupal\media_mpx\ArrayCachePoolFactory
+    arguments: ['@datetime.time', '@cache_tags.invalidator.checksum', 100]
+
+  # Use the above array cache on top of the normal HTTP request cache. We use
+  # the ChainedFastBackend cache so any invalidations on the consistent cache
+  # are shared with all threads.
+  cache.backend.media_mpx.chainedfast_http:
+    class: Drupal\Core\Cache\ChainedFastBackendFactory
+    arguments: ['@settings', null, cache.backend.media_mpx.array_cache_pool]
```
m4olivei: Here again, can we surround in single quotes and prefix with `@` for consistency?

m4olivei (commented):
I'll take you up on your walkthrough offer, I don't totally understand what I'm reading but would love to learn more.

m4olivei (approved):
Had a chat and learned some things. All is well here, great work!

PR #104, opened 2019-01-09
Update to support account context on the authenticated client

deviantintegral:
Depends on https://github.com/Lullabot/mpx-php/pull/154

TODOs:
- [x] Update composer.json when upstream has been merged.

### DIFF
--- a/src/Commands/MpxImporter.php
+++ b/src/Commands/MpxImporter.php
@@ -149,7 +149,7 @@ class MpxImporter extends DrushCommands {
 
     $factory = $this->dataObjectFactoryCreator->fromMediaSource($media_source);
     $query = new ObjectListQuery();
-    $request = $factory->selectRequest($query, $account);
+    $request = $factory->selectRequest($query);
     $list = $request->wait();
 
     // @todo Support fetching the total results via ObjectList.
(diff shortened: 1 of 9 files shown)


src/Commands/MpxImporter.php
```diff
@@ -149,7 +149,7 @@ public function warmCache(string $media_type_id) {
 
     $factory = $this->dataObjectFactoryCreator->fromMediaSource($media_source);
     $query = new ObjectListQuery();
-    $request = $factory->selectRequest($query, $account);
+    $request = $factory->selectRequest($query);
```
m4olivei, line 151: Looks like `$account` is unused in this method now. Remove definition on line 148?

m4olivei (approved):
Just one minor note. Feel free to address and merge!

deviantintegral:
Testing in combination with #105 revealed an interesting bug. We weren't actually setting an account context when listening for notifications, and were only setting it when loading the videos themselves. That led to a silently discarded 404 error, and everything _looked_ to be fine. However, adding the logging in #105 showed the errors when running cron, which amusingly led to the expected behaviour of limiting imports to the configured account.

I've added a `fromAccount` factory method to simplify this, and use it in `NotificationListener`.

m4olivei (approved):
Static review of changes looks good 👍 . Just need to address the merge conflict, then feel free to merge.

PR #105, opened 2019-01-10
Log when a notification fails (such as with a deleted video)

deviantintegral:
Without this, failures are silently discarded. I had incorrectly assumed that a `null` rejection callback would cause exceptions to bubble.


m4olivei (approved):
💯

juampynr:
@deviantintegral will this affect the migratoin by logging messages? I guess that it will. Currently when there is an error, it is printed on screen but I haven't looked at whether it is logged or not.

deviantintegral:
@juampynr this won't affect migrations as it only applies to when notifications about mpx updates are being processed. If you never run `drush mpxl` this won't apply.

PR #106, opened 2019-01-14
Default to processing all notifications at once

deviantintegral:
This adds a "reset" and "once" option to `mpx:listen`.

```
Listen for mpx notifications for a given media type.

Examples:
  media_mpx-listen mpx_video Listen for notifications for the mpx_video media type.

Arguments:
  media_type_id The media type ID to import for.

Options:
  --once  Only process a single notification.
  --reset Restarts from the earliest available notification.

Aliases: mpxl
```

By default, this means an `mpxl` call will pull in all available notifications at once. It seems stable for me at around 50MB of memory, and due to our notification filtering takes around 2 minutes for me to do a complete pull.

~~As well, I realized that means we can filter duplicate notifications across multiple notification results now. On my local, there are so many duplicate requests that processing all notifications leaves with 14 remaining. There's a small cost in that no notifications can be queued until the end, but a significant reduction in the number of entity saves we do later.~~

After more testing, I've reverted that. I'm seeing mpx block for _minutes_ (but not consistently) instead of returning a sync response after a reasonable time (say 60 seconds). That means that the queued notifications are stuck, and if the process is killed we lose them entirely.

### DIFF
--- a/src/Commands/NotificationQueuer.php
+++ b/src/Commands/NotificationQueuer.php
@@ -260,12 +260,13 @@ class NotificationQueuer extends DrushCommands {
    */
   private function doListen($media_type_id, bool $once = FALSE): void {
     $more_to_consume = TRUE;
+    $notifications_to_queue = [];
+
+    $media_type = $this->loadMediaType($media_type_id);
+    $media_source = DataObjectImporter::loadMediaSource($media_type);
 
     while ($more_to_consume) {
       // First, we find the last notification ID.
-      $media_type = $this->loadMediaType($media_type_id);
-      $media_source = DataObjectImporter::loadMediaSource($media_type);
-
       $notification_id = $this->listener->getNotificationId($media_type_id);
 
       // Next, we fetch notifications, removing duplicates (such as multiple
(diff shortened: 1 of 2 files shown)


src/Commands/NotificationQueuer.php
```diff
@@ -69,54 +69,24 @@ public function __construct(EntityTypeManagerInterface $entity_type_manager, Que
    *
    * @param string $media_type_id
    *   The media type ID to import for.
+   * @param array $options
+   *   An array of command options.
+   *
+   * @option once Only process a single notification.
```
m4olivei: I think this could be clearer. Isn't this meant to process a single _batch_ of notifications? On this line https://github.com/Lullabot/media_mpx/pull/106/files#diff-4459423a40e87d2a49044bd7b4e236a7R276 it looks like we have multiple notifications that get processed even when this param is set, it's just when it's set, it finishes with that set and returns.

m4olivei (approved):
Just one tiny suggestion, otherwise this looks good. I presume that this avoids the issue I had in my testing where I had to keep running `mpxl` manually to catch up? The robots will do that for me now?

Oh also, per CC job, the doListen is semi-big.

src/Commands/NotificationQueuer.php
deviantintegral: Yep, no more need to while true hack it!

Good thought about "once" - I really mean "notification response".

PR #107, opened 2019-01-17
Check for DateTimeFormatInterface when formatting dates

deviantintegral:
This fixes returning NULL when mapping mpx date fields to Drupal datetime fields.


m4olivei (approved):
👍

PR #108, opened 2019-01-22
Use mapped dates when calculating media availability

deviantintegral:
After a cache clear, loading an object from mpx is slower than we'd like. The most common trigger for that is checking to see if a video is available. This PR checks if the fields have been mapped, and if so uses them instead of forcing a load from mpx.

TODOs:

- [x] Update the README with this feature.

Here's a profiler comparison:

xhprof on 2019-01-22 13-55-14

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,16 @@ $factory->load('http://data.media.theplatform.com/media/data/Media/2602559', ['h
 Note that this will incur a significant performance hit on the order of 500ms
 or more, so use this option sparingly and rely on cached data where possible.
 
+## Media access and availability calculations
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -112,6 +112,16 @@ $factory->load('http://data.media.theplatform.com/media/data/Media/2602559', ['h
 Note that this will incur a significant performance hit on the order of 500ms
 or more, so use this option sparingly and rely on cached data where possible.
 
+## Media access and availability calculations
+
+Access to a mpx media entity is calculated using the mpx available and
```
juampynr: nitpick: shouldn't it be `to an mpx`?

PR #109, opened 2019-02-05
Support ingesting FeedConfigs as entities

deviantintegral:
This adds support for importing FeedConfig objects, making it straight forward for Drupal to react to upstream changes by implementing an event subscriber or an entity hook. This also saves having to refactor the Drush notification commands to support listening for updates outside of a media entity. Finally, the thumbnail download methods are pulled up to the base class so they can be used to download a reasonable thumbnail for feeds.

### DIFF
--- a/src/Plugin/media/Source/FeedConfig.php
+++ b/src/Plugin/media/Source/FeedConfig.php
@@ -57,29 +57,7 @@ class FeedConfig extends MediaSourceBase implements MediaSourceInterface {
     if (!$media->get($source_field->getName())->isEmpty()) {
       switch ($attribute_name) {
         case 'thumbnail_uri':
-          // Unlike the rest of the mpx API, these IDs are numeric and don't
-          // include the host name.
-          $factory = $this->dataObjectFactoryCreator->forObjectType($this->getAccount()->getUserEntity(), 'Media Data Service', 'Media', '1.10');
-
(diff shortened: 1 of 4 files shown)


src/Plugin/media/Source/FeedConfig.php
```diff
@@ -0,0 +1,114 @@
...
+    $feed = $this->getMpxObject($media);
+    $pinned_ids = $feed->getPinnedIds();
+    foreach ($pinned_ids as $video_id) {
+
+      try {
+        $video = $factory->loadByNumericId($video_id)->wait();
+        return $this->downloadThumbnail($video);
+      }
+      catch (ClientException $e) {
+        // Mpx doesn't removed pinned videos from feeds if the video is
```
juampynr: Shouldn't this be `doesn't remove`?

juampynr:
Static review looks fine to me. Who should review this, @deviantintegral ?

deviantintegral:
@juampynr I asked @m4olivei to comment on any PRs he had feedback on, as these have mostly been used in production for a few months now.

src/Plugin/media/Source/FeedConfig.php
deviantintegral: 🤦‍♂

PR #110, opened 2019-02-14
Fix not being able to import "0" strings and integers

deviantintegral:
This needs a test before it can be merged.


src/Plugin/media/Source/Media.php
```diff
@@ -261,25 +261,25 @@ private function getMpxMetadata(MediaInterface $media, $attribute_name) {
     $value = $this->getThumbnailMetadata($media, $attribute_name);
 
     // Check if the attribute is a core thePlatform-defined field.
-    if (!$value) {
+    if (is_null($value)) {
```
m4olivei, line 296: 💯

PR #113, opened 2019-03-05
Issue #112 Thumbnails as media entities

### LINKED ISSUE #112: Support ingesting thumbnails as media entities
Currently the mpx module ingests thumbnails as unmanaged files, following what other D8 media modules like Video Embed Field do. 

However, sometimes you want to reuse those thumbnails for use in other content like a promo for example. We need to add an option to select the media bundle (and field?) to ingest thumbnails as and to save them there, as well as exposing the final URI to the metadata thumbnail.

m4olivei:
Resolves #112.

- [x] Test the default case with all this fancy added
- [x] Tidy up config schema
- [x] Cover off cases when core and others assume the thumbnail is found in `$media->thumbnail`

### DIFF
--- a/src/DataObjectImporter.php
+++ b/src/DataObjectImporter.php
@@ -10,6 +10,7 @@ use Drupal\media\Entity\Media;
 use Drupal\media\MediaInterface;
 use Drupal\media\MediaTypeInterface;
 use Drupal\media_mpx\Event\ImportEvent;
+use Drupal\media_mpx\Plugin\media\Source\Media as MpxMediaSource;
 use Drupal\media_mpx\Plugin\media\Source\MpxMediaSourceInterface;
 use function GuzzleHttp\Psr7\build_query;
 use GuzzleHttp\Psr7\Request;
@@ -204,7 +205,6 @@ class DataObjectImporter {
    * @see https://www.drupal.org/project/drupal/issues/2878119
    */
   protected function updateThumbnail(MediaInterface $media): void {
-    /** @var \Drupal\media_mpx\Plugin\media\Source\Media $source */
     $source = $media->getSource();
     $plugin_definition = $source->getPluginDefinition();
 
@@ -223,7 +223,7 @@ class DataObjectImporter {
 
     // If the media source is configured to save the thumbnail as a media
     // entity, create the media image entity and reference the file from it.
-    if ($source->doSaveThumbnailAsMedia()) {
+    if ($source->getPluginId() === 'media_mpx_media' && $source->doSaveThumbnailAsMedia()) {
       $this->referenceThumbnailAsMedia($media, $file);
     }
   }
--- a/src/DataObjectImporter.php
+++ b/src/DataObjectImporter.php
@@ -280,19 +280,31 @@ class DataObjectImporter {
     $source = $media->getSource();
     $source_configuration = $source->getConfiguration();
 
-    // Save a media entity for the thumbnail image.
-    $media_image = Media::create([
-      'bundle' => $source_configuration['media_image_bundle'],
-      'name' => $file->label(),
-      $source_configuration['media_image_field'] => [
-        [
-          'target_id' => $file->id(),
-          'alt' => $this->getThumbnailAltForMedia($media),
-          'title' => $this->getThumbnailTitleForMedia($media),
+    // Look up whether we already have a media entity corresponding to the given
+    // file.
+    $media_storage = $this->entityTypeManager->getStorage('media');
+    $media_query = $media_storage->getQuery();
+    $existing = $media_query->condition('bundle', $source_configuration['media_image_bundle'])
+      ->condition("{$source_configuration['media_image_field']}.target_id", $file->id())
+      ->execute();
+    if ($existing) {
+      $media_image = $media_storage->load(reset($existing));
+    }
+    else {
+      // Save a media entity for the thumbnail image.
+      $media_image = Media::create([
+        'bundle' => $source_configuration['media_image_bundle'],
+        'name' => $file->label(),
+        $source_configuration['media_image_field'] => [
+          [
+            'target_id' => $file->id(),
+            'alt' => $this->getThumbnailAltForMedia($media),
+            'title' => $this->getThumbnailTitleForMedia($media),
+          ],
         ],
-      ],
-    ]);
-    $media_image->save();
+      ]);
+      $media_image->save();
+    }
     // Set a reference to the newly saved thumbnail media entity on the video
     // media entity.
     $media->{$source_configuration['media_image_entity_reference_field']} = [
--- a/src/DataObjectImporter.php
+++ b/src/DataObjectImporter.php
@@ -298,6 +298,9 @@ class DataObjectImporter {
           ],
         ],
       ]);
+      if ($owner = $media->getOwner()) {
+        $media_image->setOwner($owner);
+      }
       $media_image->save();
     }
     // Set a reference to the newly saved thumbnail media entity on the video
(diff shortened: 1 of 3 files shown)


juampynr:
This piece of logic should change so the media image entity is created, right? 

https://github.com/nbc-bravo/drupal-media_mpx/blob/patched/src/Plugin/media/Source/MediaSourceBase.php#L290

Perhaps if there is a configuration setting such as "Convert images as media entities" then the new logic is used, otherwise it does what it currently does. The problem is that then the rest of the code of this modules and modules that depend on it should take this chance into account unless they are using the right APIs in this module to obtain the thumbnail.

m4olivei:
> This piece of logic should change so the media image entity is created, right?

> https://github.com/nbc-bravo/drupal-media_mpx/blob/patched/src/Plugin/media/Source/MediaSourceBase.php#L290

I originally thought that too. However, that piece only serves to save the thumbnail to disk. Then later in the process, a file entity is created and associated to the MPX media via the `thumbnail` field: https://github.com/nbc-bravo/drupal-media_mpx/blob/patched/src/DataObjectImporter.php#L214,L233

So here I've adjusted that piece in the `DataObjectImporter`, to read the configuration to see if the media type is configured to save media entities and act accordingly: https://github.com/Lullabot/media_mpx/pull/113/files#diff-fc260f0abc8a0bba1f10e3f3917a1288R217

In both cases, the default case of saving to the `$media->thumbnail` field and this new scenario of saving to a media entity, we need a file entity, the difference is just where it's associated. In the former it's `$mpx_media->thumbnail` in the latter it's `$mpx_media->{$configured_entity_reference_field}->entity->{$configured_image_field}`.

Does that make sense?

> The problem is that then the rest of the code of this modules and modules that depend on it should take this chance into account unless they are using the right APIs in this module to obtain the thumbnail.

Yeah I still have some work to do here. For starters, I'll adjust `\Drupal\media_mpx\Plugin\media\Source\Media::getMetadata` to return the value from the right place based on the source configuration. I'm not totally sure how to handle places that go directly to the `$mpx_media->thumbnail` field, but I'm open to ideas.

src/DataObjectImporter.php
```diff
@@ -230,22 +260,107 @@ protected function updateThumbnail(MediaInterface $media): void {
...
+   *   Thumbnail file entity.
+   *
+   * @throws \Drupal\Core\Entity\EntityStorageException
+   */
+  protected function referenceThumbnailAsMedia(MediaInterface $media, FileInterface $file) {
+    $source = $media->getSource();
+    $source_configuration = $source->getConfiguration();
+
+    // Save a media entity for the thumbnail image.
+    $media_image = Media::create([
```
deviantintegral: I think we should follow what we do for files, and first see if an existing media entity exists. Otherwise, if multiple mpx assets reference the same upstream thumbnail, we could end up with duplicate media entities.

src/Plugin/media/Source/Media.php
```diff
@@ -519,4 +535,278 @@ private function getMediaValue(MediaInterface $media, $attribute_name) {
...
+    $entity = $form_state->getFormObject()->getEntity();
+
+    $form['media_image_bundle'] = [
+      '#type' => 'select',
+      '#title' => $this->t('Media image bundle'),
+      '#default_value' => $this->getConfiguration()['media_image_bundle'],
+      '#options' => $this->getImageMediaBundles(),
+      '#empty_option' => $this->t('- Select -'),
+      '#description' => $this->t('To have the thumbnail mapped to a media entity rather than a file (the default) choose a media type and field to map the thumbnail to.'),
+      '#ajax' => [
```
deviantintegral, line 495: 💯

src/Plugin/media/Source/Media.php
```diff
@@ -519,4 +535,278 @@ private function getMediaValue(MediaInterface $media, $attribute_name) {
...
+        'wrapper' => 'media-image-field',
+      ],
+    ];
+
+    if (!empty($this->getConfiguration()['media_image_bundle'])) {
+      $form['media_image_field'] = $this->getMediaImageFieldElement($this->getConfiguration()['media_image_bundle']);
+    }
+    else {
+      $form['placeholder_media_image_field'] = [
+        '#markup' => '<div id="media-image-field"></div>',
```
deviantintegral: Do we need to use `Html::UniqueId()` here?

src/Plugin/media/Source/Media.php
```diff
@@ -519,4 +535,278 @@ private function getMediaValue(MediaInterface $media, $attribute_name) {
...
+   * @return array
+   *   Array of media bundle labels keyed by bundle id, suitable for use in a
+   *   #type => select return array.
+   */
+  protected function getImageMediaBundles() {
+    // Get all media bundles.
+    $bundles = $this->entityTypeManager->getStorage('media_type')->loadMultiple();
+
+    $bundles = array_filter($bundles, function (MediaType $bundle) {
+      // Filter out any bundles that are use the same source plugin. That would
```
deviantintegral: Instead of "the same", how about "this source plugin"? I read this as multiple bundles using image, which would totally make sense.

src/Plugin/media/Source/Media.php
```diff
@@ -519,4 +535,278 @@ private function getMediaValue(MediaInterface $media, $attribute_name) {
...
+        'bundle' => $media_type->id(),
+        'label' => $this->t('Media image'),
+        'required' => FALSE,
+      ]);
+    $field->save();
+    return $field;
+  }
+
+  /**
+   * Generate a name for a new entity reference field.
```
deviantintegral, line 707: I'm surprised core doesn't have something for this, given media module has to do this when creating a new bundle.

deviantintegral (commented):
A few notes, but this PR is _so good_. Looking forward to seeing more!

src/Plugin/media/Source/Media.php
m4olivei: I don't think so. `Html::UniqueId()` will remember ids used across AJAX requests. We want an id that's consistent. 

I ended up having to play around with the AJAX stuff here a lot to get it to work. I ended up with the AJAX callback also having to return this placeholder div in the empty case. It's tricky b/c the AJAX callback doesn't seem to get the SubFormState, but the full form state, so it's unfortunately tightly coupled to the MediaType form :/.

I can prefix this id with the module name to mitigate somewhat.

src/DataObjectImporter.php
m4olivei: Great catch! I think as is an update to a video would also trigger a new media entity, which is bad 😬

src/Plugin/media/Source/Media.php
m4olivei: I know right? The MediaSourceBase has it's own protected method for doing the same. This was mostly inspired by that method: `\Drupal\media\MediaSourceBase::getSourceFieldName`.

m4olivei:
@juampynr @deviantintegral For dealing with other modules / themes directly referencing `$media->thumbnail`, what if we just always set that reference? When using media for images is configured, `$media->thumbnail` ends up being redundant. That's not a huge deal since it's just another file entity reference to the same file that the media entity references, so you end up with

`mpx_media --> thumbnail_field --> file_entity` and
`mpx_media --> configured_entity_reference_field --> media_image_entity --> configured_image_field --> file_entity`

Where `file_entity` is the same for both.

Thoughts?

juampynr:
Sounds good to me @m4olivei plus it would be backwards compatible in case there are others relying on `->thumbnail`.

deviantintegral:
👍

m4olivei:
Just merged in the latest `8.x-1.x` to resolve the merge conflict. I had a to do item here to add some test coverage for this. I'm not sure when I'll get around to that, so if someone else wants to take a stab at it, go for it.

edutrul:
Just merged to latest `8.x-1.x` the test coverage can be done in a separate PR https://github.com/Lullabot/media_mpx/issues/130 
This has been well tested in bravo and oxygen
As mentioned in slack merging this feature branch

PR #114, opened 2019-03-21
Fix migrate namespace for base class

juampynr:
MediaMpxType does not need to inherit from Paragraph's DrupalSqlBase. It can instead inherit it from migrate_drupal. This will fix errors in sites where Paragraphs is not installed.


deviantintegral:
Thanks!

Issue #120, opened 2019-05-10
Call to a member function getSource() on Null ... PlayerFormatter.php


edutrul:
I installed https://www.drupal.org/project/readonly_field_widget
and configured to `mpx_video_17` form display (I changed all fields to READ ONLY and getting below error:
![image](https://user-images.githubusercontent.com/1582129/57534187-1a9c6a80-7305-11e9-9782-c4d79ccafc3b.png)

I don't think the issue may come directly to readonly_field_widget.
My proposal fix is to add defensive programming(make sure the entity object exists!)

m4olivei:
@edutrul Try @salvamomo's suggestion from https://github.com/nbc-bravo/drupal-media_mpx/pull/21#discussion_r282913958. I believe that will lower the cognitive complexity as code climate sees it. I also agree with @salvamomo that it's a nice approach for cleaner diff.

deviantintegral:
Looks like this is done!

PR #121, opened 2019-05-10
#120 Add defensive programming

edutrul:
Original issue: https://github.com/Lullabot/media_mpx/issues/120

I installed https://www.drupal.org/project/readonly_field_widget
and configured to `mpx_video_17` form display (I changed all fields to READ ONLY and getting below error:
image

I don't think the issue may come directly to readonly_field_widget.
My proposal fix is to add defensive programming(make sure the entity object exists!)

### DIFF
--- a/src/Plugin/Field/FieldFormatter/PlayerFormatter.php
+++ b/src/Plugin/Field/FieldFormatter/PlayerFormatter.php
@@ -127,21 +127,23 @@ class PlayerFormatter extends FormatterBase implements ContainerFactoryPluginInt
 
     /** @var \Drupal\media\Entity\Media $entity */
     $entity = $items->getEntity();
-    /** @var \Drupal\media_mpx\Plugin\media\Source\Media $source_plugin */
-    $source_plugin = $entity->getSource();
+    if ($entity) {
+      /** @var \Drupal\media_mpx\Plugin\media\Source\Media $source_plugin */
+      $source_plugin = $entity->getSource();
 
-    // @todo Cache this.
-    $factory = $this->dataObjectFactoryCreator->forObjectType($source_plugin->getAccount()->getUserEntity(), 'Player Data Service', 'Player', '1.6');
+      // @todo Cache this.
+      $factory = $this->dataObjectFactoryCreator->forObjectType($source_plugin->getAccount()->getUserEntity(), 'Player Data Service', 'Player', '1.6');
 
-    try {
-      $player = $factory->load(new Uri($this->getSetting('player')))->wait();
+      try {
+        $player = $factory->load(new Uri($this->getSetting('player')))->wait();
+      }
+      catch (TransferException $e) {
+        // If we can't load a player, we can't render any elements.
+        $this->mpxLogger->logException($e);
+        return $element;
+      }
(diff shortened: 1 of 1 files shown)


edutrul:
Code climate is failing at:
Function fetchPlayerOptions has a Cognitive Complexity of 6 (exceeds 5 allowed). Consider refactoring. NEW
```
  protected function fetchPlayerOptions(): array {
    $options = [];
    $bundle = $this->fieldDefinition->getTargetBundle();
    if ($bundle) {
      /** @var \Drupal\media\Entity\MediaType $type */
``` 
image
I am not sure about cognitive complexity. 

I can move this to another function but prefer to keep in the same function fetchPlayerOptions for readability purposes. Is there a way to ignore this error or with special comments to ignore codeClimate cognitive complexity @m4olivei @deviantintegral ?

src/Plugin/Field/FieldFormatter/PlayerFormatter.php
```diff
@@ -228,23 +230,25 @@ public static function defaultSettings() {
...
+      $factory = $this->dataObjectFactoryCreator->forObjectType($source_plugin->getAccount()->getUserEntity(), 'Player Data Service', 'Player', '1.6', $source_plugin->getAccount());
+      $query = new ObjectListQuery();
+      $sort = new Sort();
+      $sort->addSort('title');
+      $query->setSort($sort);
+
+      /** @var \Lullabot\Mpx\DataService\Player\Player[] $results */
+      $results = $factory->select($query);
+
+      foreach ($results as $player) {
```
salvamomo: If each result is a player option, shouldn't it be called `$player_option`, or `$option`?

salvamomo (commented):
@edutrul: Leaving my response here for the question about cognitive complexity. It's related to what I mentioned in the PR opened for the nbc-bravo repo: https://github.com/nbc-bravo/drupal-media_mpx/pull/21/files#r282913958. It looks at things like number of conditionals in a method, switch statements, loops, nesting of items, etc. 

For this particular case, if you apply the feedback I left in the link above and return early if there is no $bundle, instead of nesting the whole logic inside that `if ($bundle)` check, the method automatically becomes simpler to read, (plus you get to have more space in all the lines that follow). I believe just that change would probably make code climate to pass 👍 . 

For even better readability, you could consider extracting a lot of the logic into a private method `fetchPlayerOptionsForAccount`, that receives a `@var \Drupal\media_mpx\Entity\Account`, and replace everything from the `$factory = $this... ` line, with a call to the other function. That makes it clear that the first one holds the logic to figure out "what account belongs to the given bundle", and the second method's name makes it self-explanatory with the name that it fetches the player options for a given account, without having to worry about the bundle. Take this just as general feedback, I don't think it's blocking this from merging :). My only question is the same I left here: https://github.com/nbc-bravo/drupal-media_mpx/pull/21#issuecomment-491315584. If this approach is fine, then all is good 👍 

Further info on the CC from climate: https://docs.codeclimate.com/docs/cognitive-complexity.

salvamomo (approved):

m4olivei (approved):

PR #122, opened 2019-05-14
Queue existing content

salvamomo:
# Description
This PR adds an admin UI (with a dedicated permission) to allow administrators queue mpx media type imports. The queueing process is done in batches.

There's a minor section of code for which a change is required in the mpx-php library. Linked PR: https://github.com/Lullabot/mpx-php/pull/158.

```
    $request = new QueueVideoImportsRequest($media_type_id,
      $batch_size,
      $context['sandbox']['last_index'] + 1);

    $response = $service->execute($request);

    if (!isset($context['sandbox']['total_results'])) {
      $context['sandbox']['total_results'] = $response->getIterator()->getList()->getTotalResults();
    }
```

# Screenshots:
### Form UI.
queue_1

### Batch process running.
queue_2

### New permission.
queue_permissions

### Batch process finished.
queue_background


media_mpx.routing.yml
```diff
@@ -37,3 +37,23 @@ media_mpx.settings:
     _title: 'mpx settings'
   requirements:
     _permission: 'administer site configuration'
+
+media_mpx.asset_sync:
+  path: '/admin/config/media/mpx/asset-sync'
```
deviantintegral, line 42: Do you think these should be under config, or a subtask under content/media?

src/Exception/MediaTypeDoesNotExistException.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Drupal\media_mpx\Exception;
+
+/**
+ * Class MediaTypeDoesNotExistException.
+ *
+ * @package Drupal\media_mpx\Exception
+ */
+class MediaTypeDoesNotExistException extends \Exception {}
```
deviantintegral: Should this extend `\Drupal\Core\Entity\EntityStorageException`? Or if not, perhaps `\RuntimeException`?

src/Exception/MediaTypeNotAssociatedWithMpxException.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Drupal\media_mpx\Exception;
+
+/**
+ * Class MediaTypeNotAssociatedWithMpxException.
+ *
+ * @package Drupal\media_mpx\Exception
+ */
+class MediaTypeNotAssociatedWithMpxException extends \Exception {}
```
deviantintegral: Same question, should this extend some other exception type? At a glance it almost feels like `\LogicException`, or perhaps `\InvalidArgumentException`.

src/Form/QueueContentsForm.php
```diff
@@ -0,0 +1,173 @@
...
+
+    $form['mpx_video_types'] = [
+      '#type' => 'checkboxes',
+      '#title' => t('mpx Video Types'),
+      '#options' => $options,
+      '#required' => TRUE,
+    ];
+    $form['queue'] = [
+      '#type' => 'submit',
+      '#value' => $this->t('Queue selected Video Types'),
```
deviantintegral: We're not queuing the _types_ specifically, but the videos associated with each type. How about just `Import videos`?

src/Form/QueueContentsForm.php
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    $types = array_filter(array_values($form_state->getValue('mpx_video_types')));
+
+    $batch_builder = new BatchBuilder();
+    $batch_builder->setTitle(t('MPX Video Imports Queueing'))
```
deviantintegral: `Finding mpx videos to import`?

src/Form/QueueContentsForm.php
```diff
@@ -0,0 +1,173 @@
...
+      ->setFinishCallback([$this, 'batchFinished'])
+      ->setInitMessage(t('Starting Queueing Process.'));
```
deviantintegral: `Starting import`?

src/Form/QueueContentsForm.php
```diff
@@ -0,0 +1,173 @@
...
+   *   Context data passed between batches.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   * @throws \Drupal\media_mpx\Exception\MediaTypeDoesNotExistException
+   * @throws \Drupal\media_mpx\Exception\MediaTypeNotAssociatedWithMpxException
+   */
+  public function batchedQueueItemsForMediaType(string $media_type_id, array &$context) {
+    /* @var \Drupal\media_mpx\Service\QueueVideoImports $service */
+    $service = \Drupal::getContainer()->get('media_mpx.service.queue_video_imports');
```
deviantintegral: Does this have to go to `\Drupal`? Asking because you use constructor injection earlier.

src/Form/QueueContentsForm.php
```diff
@@ -0,0 +1,173 @@
...
+    $batch_size = 200;
```
deviantintegral, line 102: We should document how you came up with this number.

src/Form/QueueContentsForm.php
```diff
@@ -0,0 +1,173 @@
...
+      $context['sandbox']['total_results'] = $response->getIterator()->getList()->getTotalResults();
+    }
+    $total_items = $context['sandbox']['total_results'];
+
+    $context['results']['success'] += $response->getVideosQueued();
+    $context['results']['errors'] += $response->getErrors();
+    $context['sandbox']['progress'] += $response->getVideosQueued();
+    $context['sandbox']['last_index'] = $context['sandbox']['last_index'] + $batch_size;
+
+    $context['message'] = 'Processed a total of ' . $context['sandbox']['progress'] . ' items.';
```
deviantintegral: This array assignment is ignored in the line below?

src/Form/QueueContentsForm.php
```diff
@@ -0,0 +1,173 @@
...
+  /**
+   * Finished callback for the batch queuing operations to update mpx videos.
+   *
+   * @param bool $success
+   *   Result of the queuing operations.
+   * @param array $results
+   *   An array with the count of successful and errored queue operations.
+   */
+  public function batchFinished(bool $success, array $results) {
+    $finished_message = t('Finished with errors.');
```
deviantintegral: `\t` here? Or `$this->t()`?

src/Form/QueueContentsForm.php
```diff
@@ -0,0 +1,173 @@
...
+    $messenger = \Drupal::messenger();
```
deviantintegral: I believe there's `$this->messenger()` on the base class you can use.

src/Form/QueueContentsForm.php
```diff
@@ -0,0 +1,173 @@
...
+
+    if ($success) {
+      $finished_message = \Drupal::translation()->formatPlural(
+        $results['success'],
+        'One mpx item queue.',
```
deviantintegral: `One mpx item queued.`?

src/Form/QueueContentsForm.php
```diff
@@ -0,0 +1,173 @@
...
+        '@count mpx items queued for updates.');
+    }
+
+    $messenger->addMessage($finished_message);
+
+    if ($results['errors'] > 0) {
+      $errors_message = \Drupal::translation()->formatPlural(
```
deviantintegral: `$this->formatPlural()`?

src/Form/QueueContentsForm.php
```diff
@@ -0,0 +1,173 @@
...
+        $results['errors'],
+        'One mpx item could not be queued (check logs for details).',
```
deviantintegral: `One mpx item could not be queued. See the site logs for details.` -> saves the parenthesis for what are two sentences.

src/Form/QueueContentsForm.php
```diff
@@ -0,0 +1,173 @@
...
+        '@count mpx items could not be queued for updates (check logs for details).');
```
deviantintegral: Same thing here.

src/Form/QueueContentsForm.php
```diff
@@ -0,0 +1,173 @@
...
+   *   Result of the queuing operations.
+   * @param array $results
+   *   An array with the count of successful and errored queue operations.
+   */
+  public function batchFinished(bool $success, array $results) {
+    $finished_message = t('Finished with errors.');
+    $messenger = \Drupal::messenger();
+
+    if ($success) {
+      $finished_message = \Drupal::translation()->formatPlural(
```
deviantintegral: `$this->formatPlural()` here.

src/Form/QueueContentsForm.php
deviantintegral: I think we should clarify that videos will be updated in the background here (and in the plural message too). Something like:

`@count mpx items queued for updates. Imports will continue in the background.`

src/Form/QueueContentsForm.php
```diff
@@ -0,0 +1,173 @@
...
+   * {@inheritdoc}
+   */
+  public function getFormId() {
+    return 'media_mpx_asset_sync_queue';
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container) {
```
deviantintegral: A good convention is to place create right after the constructor. That way, if arguments change this method is easy to see that it exists.

src/Repository/MpxMediaType.php
```diff
@@ -0,0 +1,90 @@
...
+use Drupal\media_mpx\Exception\MediaTypeDoesNotExistException;
+use Drupal\media_mpx\Exception\MediaTypeNotAssociatedWithMpxException;
+use Drupal\media_mpx\Plugin\media\Source\Media;
+
+/**
+ * Class MpxMediaType.
+ *
+ * @package Drupal\media_mpx\Repository
+ */
+class MpxMediaType {
```
deviantintegral, line 18: 👍 - i bet there's lots of other places in the code we could reuse this.

src/Service/QueueVideoImports.php
```diff
@@ -0,0 +1,156 @@
...
+use Drupal\media_mpx\MpxImportTask;
+use Drupal\media_mpx\Repository\MpxMediaType;
+use Lullabot\Mpx\DataService\Fields;
+use Lullabot\Mpx\DataService\ObjectListIterator;
+use Lullabot\Mpx\DataService\ObjectListQuery;
+use Lullabot\Mpx\DataService\Range;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+
+/**
+ * Class QueueVideoImports.
```
deviantintegral: Can we describe what this class does?

src/Service/QueueVideoImports.php
```diff
@@ -0,0 +1,156 @@
...
+ *
+ * @package Drupal\media_mpx\Service
+ */
+class QueueVideoImports {
+
+  const MEDIA_MPX_IMPORT_QUEUE = 'media_mpx_importer';
```
deviantintegral, line 27: Should we remove the constant at `\Drupal\media_mpx\Commands\MpxImporter::MEDIA_MPX_IMPORT_QUEUE` to use this one?

deviantintegral (commented):
Lots of little notes, but this is looking very nice!

Do you think we can remove some of the code in `\Drupal\media_mpx\Commands\MpxImporter`? It'd be good to know that as much of the code as possible is shared between using Drush and using the UI.

When you're done this round of edits, some screenshots of the new forms would be 💯 .

src/Repository/MpxMediaType.php
salvamomo: Totally. I'll keep an eye in the upcoming tasks for further consolidation.

src/Form/QueueContentsForm.php
salvamomo: 🤦‍♂ . I thought I couldn't do this because being callbacks generated on the fly from the DB information, the class doesn't keep any of the actual dependencies passed in construct (or above layers). Rather conveniently this doesn't affect the translation service (nor the `\Drupal::messenger()` above which I just cleaned up as well) because that's provided statically from a Trait, instead of via DI

media_mpx.routing.yml
salvamomo: That's a good point. Given this is a contrib and we don't know how bloated media can be already, maybe a dedicated area? https://monosnap.com/file/ovTWNU9MUYwUUyYYurEddXzvnxQDRv

I've no hard preference either way. If that's ok I'll tackle the moving of this in the next task, given there are more UI items to come.

src/Exception/MediaTypeDoesNotExistException.php
salvamomo: Changed 👍

src/Exception/MediaTypeNotAssociatedWithMpxException.php
salvamomo: Moved to \Logic one. Great call :)

src/Form/QueueContentsForm.php
salvamomo: Yeah, clearly English my is brightest not today at its. Changed!

src/Service/QueueVideoImports.php
salvamomo: @deviantintegral I'd leave this one as is for now. Since I'll try to make the importer reuse the queue service, I should be able to get rid of the one in from `\Drupal\media_mpx\Commands\MpxImporter`.

src/Service/QueueVideoImports.php
salvamomo: Changed to `Queues MpxImportTask items to be processed by background processes.`

src/Form/QueueContentsForm.php
salvamomo: Changed to "Queueing mpx videos for background imports.", just to clear any possible confusion about the import actually happening at that exact moment.

src/Form/QueueContentsForm.php
salvamomo: @deviantintegral sadly yes, for what I've been able to test. Drupal's batch API is still a bit archaic, so when storing the operation info, we're losing the dependencies injected in the form.

src/Form/QueueContentsForm.php
salvamomo: I spotted this by chance when doing clean up, so no longer needed, but yes, it was being ignored :)

src/Form/QueueContentsForm.php
salvamomo: Added:

```
    // Mpx responses might be a bit heavy, so limit batches size to 200. Don't
    // want php running out of memory.
    // @see vendor/lullabot/mpx-php/src/DataService/ObjectListQuery.php
```

src/Form/QueueContentsForm.php
salvamomo: I think I was fixing this at the time you were reporting it :D

src/Form/QueueContentsForm.php
salvamomo: Yes! I also clarified it in the init message 👍

salvamomo:
@deviantintegral All changes applied, except a couple comments about refactoring that I plan to look into as part of the other tasks (refactoring importer command to use the new service).

Also, I think we're good to ignore the remaining code climate reports.

media_mpx.routing.yml
deviantintegral: In D7, it's under Media -> mpx, I think it makes sense as a subtask there? We can always move it later. Either way config is better than config since most editors wouldn't see config sections of admin.

src/Form/QueueContentsForm.php
deviantintegral: 😧

src/Service/QueueVideoImports.php
```diff
@@ -0,0 +1,174 @@
...
+
+      if ($this->queueMpxItem($mpx_media, $media_type->id()) == TRUE) {
+        $queued++;
+      }
+      else {
+        $errored++;
+      }
+    }
+
+    return new QueueVideoImportsResponse($queued, $errored, $results);
```
deviantintegral, line 111: I see a few ways to reduce the complexity of this function, but I actually think this class is the root of things.

The problem with this class is that we're storing _counts_ of `$queued` and `$errored`, but losing track of which specific mpx items passed or failed. As a site admin, if something fails I generally want to know which video failed so I can try to fix it. If we were not sending in `$results` this would make sense, because we could potentially let the GC free those mpx media objects, but we're still keeping a reference around.

If `queueMpxItem` returns an object, we could have code that looks something like:

```php
$limit = min($limit, count($results));
for ($index = 0; $index < $limit; $index++) {
  $mpx_media = $results[$index];
  $queue_results[] = $this->queueMpxItem($mpx_media, $media_type->id());
}
return new QueueVideoImportsResponse($queue_results);
```

I don't think we need to hold this PR on this, but let's get a followup done before closing our ticket.

PR #123, opened 2019-05-15
Improve QueueVideoImports response objects with more meaningful data

salvamomo:
## Description
- Implements feedback from https://github.com/Lullabot/media_mpx/pull/122#discussion_r284058122, although with some minor differences in regards to response handling, given the suggested snippet wasn't valid with the current `ObjectListIterator` class (e.g: calling count() on the $results array).
- Additionally to that refactor, I've leveraged it in the admin ui. Screenshot below:
- Adds unit tests for the new `QueueMpxImportResult` class.

--- 

queue_errors

EDIT: Confirming that code climate is not complaining anymore about the refactored service 🎉  (not that it's silent, but that it actually reports a "Fixed" message for the original smell reported).

### DIFF
--- a/src/Form/QueueContentsForm.php
+++ b/src/Form/QueueContentsForm.php
@@ -105,8 +105,8 @@ class QueueContentsForm extends FormBase {
       $context['sandbox']['progress'] = 0;
       $context['sandbox']['last_index'] = 0;
       $context['results'] = [
-        'success' => 0,
-        'errors' => 0,
+        'success' => [],
+        'errors' => [],
       ];
     }
 
@@ -136,9 +136,9 @@ class QueueContentsForm extends FormBase {
     $context['sandbox']['total_results'] = $context['sandbox']['total_results'] ?: $response->getIterator()->getTotalResults();
     $total_items = $context['sandbox']['total_results'];
 
-    $context['results']['success'] += $response->getVideosQueued();
-    $context['results']['errors'] += $response->getErrors();
-    $context['sandbox']['progress'] += $response->getVideosQueued();
+    $context['results']['success'] = array_merge($context['results']['success'], $response->getQueuedVideos());
+    $context['results']['errors'] = array_merge($context['results']['errors'], $response->getNotQueuedVideos());
+    $context['sandbox']['progress'] += count($response->getQueuedVideos());
     $context['sandbox']['last_index'] = $context['sandbox']['last_index'] + $batch_size;
     $context['message'] = $this->t('Queued @progress %type items out of @total.', [
       '@progress' => $context['sandbox']['progress'],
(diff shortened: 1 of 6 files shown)


src/Form/QueueContentsForm.php
```diff
@@ -161,19 +161,24 @@ public function batchFinished(bool $success, array $results) {
...
-        'One mpx item could not be queued. Check logs for details.',
-        '@count mpx items could not be queued for updates. Check logs for details.');
+        count($results['errors']),
+        'The following mpx item could not be queued. Check logs for details.',
+        'The following @count mpx items could not be queued for updates. Check logs for details.');
       $this->messenger()->addError($errors_message);
+
+      /* @var \Lullabot\Mpx\DataService\Media\Media $mpx_item */
+      foreach ($results['errors'] as $mpx_item) {
+        $this->messenger()->addError($mpx_item->getId());
```
deviantintegral, line 180: I wonder if we need to do this all as a single addError call. Otherwise, if someone has multiple tabs open, the session messages could be consumed and displayed without the first sentence as context.

src/Service/QueueMpxImportResult.php
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * QueueMpxImportResult constructor.
+   *
+   * @param bool $success
+   *   Whether the queueing attempt was successful or not.
+   * @param \Lullabot\Mpx\DataService\Media\Media $mpx_media
+   *   The mpx Media item.
+   */
+  public function __construct(bool $success, Media $mpx_media) {
```
deviantintegral: What do you think about flipping this, so the constructor is "subject, status" instead of "status, subject".

deviantintegral:
Screen Shot 2019-05-15 at 4 40 47 PM


🎉

src/Form/QueueContentsForm.php
salvamomo: @deviantintegral I'm fine with doing it in a single call, but I'm not sure that scenario is feasible at all. Keep in mind this is done in the "finished" callback and it's called just once at the very end, not in the operations one between batches.

tbc:
Was I mentioned as a typo? Maybe another dev should be notified?

salvamomo:
@tbc ouch, that's a funny one. I opened this with an empty description "@ tbc", which I use as a ("@ todo"), so CI tasks are triggered while I write. It seems github was eager to notify someone! Sorry about that :)

src/Service/QueueMpxImportResult.php
salvamomo: Done

salvamomo:
Flipped constructor arguments. Not touched the message generation yet, for the reason mentioned above.

deviantintegral (approved):

PR #124, opened 2019-05-16
Update asset by account guid

salvamomo:
## Description
- Adds a new administration UI for content editors to allow them update a given video item using its `guid` field, for a selected account.
- Moves the existing **Queue items for import** UI into admin/content/media, as a sub task (old location was under `/admin/config/....`.
  - Refactored MpxImporterQueueWorker to use the new `UpdateVideoItem` service, as that's where the  core domain logic for updating an item is isolated from context-specific updates (ui, cron, command line, etc).

- Remaining work.
  - TESTS: Not tackling here. There's a separate task that also deals with UI for updates, so I will include them there, in case any further refactoring takes place.

---

image


src/Service/UpdateVideoItem/UpdateVideoItemRequest.php
```diff
@@ -0,0 +1,83 @@
...
+   *
+   * @param \GuzzleHttp\Psr7\Uri $mpxId
+   *   The Uri object specifying the video item id.
+   * @param string $mediaTypeId
+   *   The media type id.
+   */
+  private function __construct(Uri $mpxId, string $mediaTypeId) {
+    $this->mpxId = $mpxId;
+    $this->mediaTypeId = $mediaTypeId;
+  }
```
salvamomo, line 42: constructor is private so that it's clear callers need to use one of the existing factory methods. For now there's just one, but as I'm refactoring the Importer, that'll need its own constructor, as it provides the data directly with the mpx id and the media type id.

deviantintegral: _what_ you can make private constructors!!??! 💥

src/Form/UpdateMediaItemForAccount.php
```diff
@@ -0,0 +1,170 @@
...
+
+  /**
+   * Submit handler for the 'media_mpx_asset_sync_single_by_account_guid' form.
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    $account = $form_state->getValue('account');
+    $guid = $form_state->getValue('guid');
+
+    if (!$media_items = $this->loadMediaEntitiesWithMatchingGuid($guid)) {
+      $this->messenger()->addError($this->t('There are no video items with the entered GUID.'));
```
deviantintegral: What if I'm importing a new video?

Actually, that makes me think I've suggested the wrong approach here - instead of selecting an account, it should be a media bundle?

src/Form/UpdateMediaItemForAccount.php
```diff
@@ -0,0 +1,170 @@
...
+   *
+   * @param string $guid
+   *   The guid to filter by.
+   *
+   * @return \Drupal\media\Entity\Media[]
+   *   The mpx Media entities matching the given guid
+   */
+  private function loadMediaEntitiesWithMatchingGuid(string $guid): array {
+    $media_ids = $this->mediaStorage->getQuery()
+      ->condition('field_mpx_guid', $guid)
```
deviantintegral: This field is defined by sites, so the name may be different. I think this would need to be `getMetadata(...)` from the media source plugin.

src/Service/UpdateVideoItem/UpdateVideoItem.php
```diff
@@ -0,0 +1,83 @@
...
+   * @throws \Drupal\media_mpx\Exception\MediaTypeDoesNotExistException
+   * @throws \Drupal\media_mpx\Exception\MediaTypeNotAssociatedWithMpxException
+   */
+  public function execute(UpdateVideoItemRequest $request): UpdateVideoItemResponse {
+    $mpx_id = $request->getMpxId();
+    $media_type_id = $request->getMediaTypeId();
+
+    $media_type = $this->mpxMediaTypeRepository->findByTypeId($media_type_id);
+
+    $media_source = $this->importer::loadMediaSource($media_type);
```
deviantintegral: why `::` here?

src/Service/UpdateVideoItem/UpdateVideoItemRequest.php
```diff
@@ -0,0 +1,83 @@
...
+use GuzzleHttp\Psr7\Uri;
+
+/**
+ * Contains the data needed to update a media entity via UpdateVideoItem.
+ *
+ * @package Drupal\media_mpx\Service\UpdateVideoItem
+ */
+class UpdateVideoItemRequest {
+
+  const MPX_MEDIA_DATA_SERVICE_URL = "http://data.media.theplatform.com/media/data/Media/";
```
deviantintegral: This is not actually a constant, and there's an API for this as the hostname may change. However the given id for an mpx object will be static once it's set.

Actually, I'm not sure this is needed. You're concatenating this to the full id below, which I think will duplicate this portion.

deviantintegral (commented):
Nice work! This is well on the way, a few notes I've left that probably won't be too much to deal with. Other than wrong requirements! 🤦‍♂

src/Service/UpdateVideoItem/UpdateVideoItem.php
salvamomo: Ah great catch! Leftover from its usage on the Queue Worker, which wasn't using a service, but the class directly 👍

salvamomo: tackled.

src/Service/UpdateVideoItem/UpdateVideoItemRequest.php
salvamomo: php is the new java 😆

src/Form/UpdateMediaItemForAccount.php
salvamomo: Right, totally. I actually wondered why this was for updating, and not for both creation and updates. I'll rework this! 🚀

src/Service/UpdateVideoItem/UpdateVideoItemRequest.php
salvamomo: It won't duplicate it, as I'm using ->load() which expects the Uri already.... Buuut I just noticed there's a `loadByNumericId` method, which indeed makes this oddity here totally unnecessary. Fixing 👍

src/Form/UpdateMediaItemForAccount.php
salvamomo: 2 hours of mpx pain later: This can't be done. Details in ticket, but essentially retrieving object data by **guid** is not working for me. So we'll have to restrict this UI to **update only** ops, and defer to the upcoming _global id ui_ for both create/update imports.

src/Form/UpdateMediaItemForAccount.php
salvamomo: It's taken a while, as I thought this was configured, and went around in circles wondering why fieldMap was coming with empty results all the time. oh well. 🤦‍♂  Done now :).

salvamomo:
Changes and clean up done. This should be ready.

src/Form/UpdateMediaItemForVideoType.php
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    $updateRequest = UpdateVideoItemRequest::createFromMediaEntity($video_item);
+    try {
+      $this->updateVideoItemService->execute($updateRequest);
+      $this->messenger()->addMessage($this->t('The selected video has been updated.'));
+    }
+    catch (\Exception $e) {
+      // Up until here, all necessary checks have been made. No custom exception
+      // handling needed other than for the db possibly exploding at this point.
```
juampynr, line 126: :boom: :smile:

src/Service/QueueVideoImports.php
```diff
@@ -96,10 +96,13 @@ public function execute(QueueVideoImportsRequest $request): QueueVideoImportsRes
     $queue_results = [];
 
     foreach ($results as $index => $mpx_media) {
+      $queue_results[] = $this->queueMpxItem($mpx_media, $media_type->id());
+
+      // Break if the limist is reached, to avoid an extra request to mpx for
```
juampynr, line 101: Limit.

Can be fixed in a straight push to master.

juampynr:
Looks good! Merging.

src/Service/QueueVideoImports.php
salvamomo: oops! Will fix 👍

PR #126, opened 2019-05-20
Update asset by mpx

salvamomo:
## Summary of Changes
- Adds a new form under `admin/content/media/mpx/update-item-by-mpx-id` that allows content managers / administrators to import (create or update) a video item by entering its `mpx ID`. Video type dropdown is required because it's needed to know mappings for the item, etc.

- Adds Kernel tests for the `UpdateVideoItem` service.

### DIFF
--- a/src/Form/ImportVideoItemByMpxId.php
+++ b/src/Form/ImportVideoItemByMpxId.php
@@ -19,7 +19,7 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  *
  * @package Drupal\media_mpx\Form
  */
-class ImportMediaItemByMpxId extends FormBase {
+class ImportVideoItemByMpxId extends FormBase {
 
   /**
    * The Update Video Item service.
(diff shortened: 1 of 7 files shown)


src/Form/ImportVideoItemByMpxId.php
```diff
@@ -0,0 +1,165 @@
...
+      $container->get('logger.channel.media_mpx')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    if (!$video_opts = $this->loadVideoTypeOptions()) {
+      $this->messenger()->addError($this->t('There has been an unexpected problem loading the form. Reload the page.'));
```
juampynr, line 70: @salvamomo why could this happen? Just curious. Also, would reloading the page solve it?

salvamomo: Rather unlikely, and not related this module itself (say db connection failing for a moment, which could break the whole site anyway and not only this page), but not impossible.

deviantintegral: This is fine for now, but a more robust method would be to throw an exception from `loadVideoTypeOptions()`. "object || false" returns are evil :D

PR #127, opened 2019-05-20
Add button to update media item from its edit form

salvamomo:
## Summary 

- Adds a "Update with mpx data" button to the mpx media forms. This will trigger the update service to get that item updated.
- Some clean up of duplicated code around exception logging.

### Remaining tasks
- [x] UI messages for success / error.
- [x] Error logging.

### Screenshots
media_mpx_reimport

**Attempt to reimport before any mapping is actually configured**
media_mpx_configure_media_source

### DIFF
--- a/media_mpx.module
+++ b/media_mpx.module
@@ -185,6 +185,10 @@ function media_mpx_migration_plugins_alter(array &$migrations) {
  * Implements hook_form_FORM_ID_alter().
  */
 function media_mpx_form_media_form_alter(&$form, $form_state, $form_id) {
-  $media_form_alter = MediaFormAlter::create(\Drupal::getContainer());
+  $mpx_type_repository = \Drupal::service('media_mpx.repository.mpx_media_types');
+  $update_service = \Drupal::service('media_mpx.service.update_video_item');
+  $logger = \Drupal::service('media_mpx.exception_logger');
+
+  $media_form_alter = new MediaFormAlter($mpx_type_repository, $update_service, $logger);
   $media_form_alter->alter($form, $form_state, $form_id);
 }
--- a/media_mpx.module
+++ b/media_mpx.module
@@ -11,6 +11,7 @@ use Drupal\Core\Routing\RouteMatchInterface;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\Url;
 use Drupal\media\MediaInterface;
+use Drupal\media_mpx\FormAlter\MediaFormAlter;
 use Drupal\media_mpx\Plugin\QueueWorker\ThumbnailDownloader;
 
 /**
@@ -178,4 +179,12 @@ function media_mpx_migration_plugins_alter(array &$migrations) {
       ];
     }
   }
+}
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function media_mpx_form_media_form_alter(&$form, $form_state, $form_id) {
+  $media_form_alter = MediaFormAlter::create(\Drupal::getContainer());
+  $media_form_alter->alter($form, $form_state, $form_id);
 }
--- a/media_mpx.module
+++ b/media_mpx.module
@@ -185,10 +185,9 @@ function media_mpx_migration_plugins_alter(array &$migrations) {
  * Implements hook_form_FORM_ID_alter().
  */
 function media_mpx_form_media_form_alter(&$form, $form_state, $form_id) {
-  $mpx_type_repository = \Drupal::service('media_mpx.repository.mpx_media_types');
   $update_service = \Drupal::service('media_mpx.service.update_video_item');
   $logger = \Drupal::service('media_mpx.exception_logger');
 
-  $media_form_alter = new MediaFormAlter($mpx_type_repository, $update_service, $logger);
+  $media_form_alter = new MediaFormAlter($update_service, $logger);
   $media_form_alter->alter($form, $form_state, $form_id);
 }
--- a/src/Form/UpdateMediaItemForVideoType.php
+++ b/src/Form/UpdateMediaItemForVideoType.php
@@ -7,13 +7,11 @@ namespace Drupal\media_mpx\Form;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Form\FormBase;
 use Drupal\Core\Form\FormStateInterface;
-use Drupal\Core\Logger\RfcLogLevel;
-use Drupal\Core\Utility\Error;
 use Drupal\media\Entity\Media;
+use Drupal\media_mpx\MpxLogger;
 use Drupal\media_mpx\Repository\MpxMediaType;
 use Drupal\media_mpx\Service\UpdateVideoItem\UpdateVideoItem;
 use Drupal\media_mpx\Service\UpdateVideoItem\UpdateVideoItemRequest;
-use Psr\Log\LoggerInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
@@ -45,16 +43,16 @@ class UpdateMediaItemForVideoType extends FormBase {
   private $mpxTypeRepository;
 
   /**
-   * The system logger.
+   * The custom media mpx logger.
    *
-   * @var \Psr\Log\LoggerInterface
+   * @var \Drupal\media_mpx\MpxLogger
    */
   private $logger;
 
   /**
    * UpdateMediaItemForAccount constructor.
    */
-  public function __construct(EntityTypeManagerInterface $entityTypeManager, UpdateVideoItem $updateVideoItem, MpxMediaType $mpxTypeRepository, LoggerInterface $logger) {
+  public function __construct(EntityTypeManagerInterface $entityTypeManager, UpdateVideoItem $updateVideoItem, MpxMediaType $mpxTypeRepository, MpxLogger $logger) {
     $this->mediaStorage = $entityTypeManager->getStorage('media');
     $this->updateVideoItemService = $updateVideoItem;
     $this->mpxTypeRepository = $mpxTypeRepository;
@@ -69,7 +67,7 @@ class UpdateMediaItemForVideoType extends FormBase {
       $container->get('entity_type.manager'),
       $container->get('media_mpx.service.update_video_item'),
       $container->get('media_mpx.repository.mpx_media_types'),
-      $container->get('logger.channel.media_mpx')
+      $container->get('media_mpx.exception_logger')
     );
   }
 
@@ -125,7 +123,7 @@ class UpdateMediaItemForVideoType extends FormBase {
       // Up until here, all necessary checks have been made. No custom exception
       // handling needed other than for the db possibly exploding at this point.
       $this->messenger()->addError($this->t('There has been an unexpected problem updating the video. Check the logs for details.'));
-      $this->watchdogException($e);
+      $this->logger->watchdogException($e, sprintf('mpx video with guid %s could not be updated', $guid));
     }
   }
 
@@ -172,26 +170,6 @@ class UpdateMediaItemForVideoType extends FormBase {
   }
 
   /**
-   * Logs an exception.
-   *
-   * @param \Exception $exception
-   *   The exception that is going to be logged.
-   * @param string $message
-   *   The message to store in the log.
-   *
-   * @see \Drupal\Core\Utility\Error::decodeException()
-   */
-  private function watchdogException(\Exception $exception, $message = NULL) {
-    // Use a default value if $message is not set.
-    if (empty($message)) {
-      $message = '%type: @message in %function (line %line of %file).';
-    }
-
-    $variables = Error::decodeException($exception);
-    $this->logger->log(RfcLogLevel::ERROR, $message, $variables);
-  }
-
-  /**
    * Returns the mpx Video Type options of the dropdown (prepared for form api).
    *
    * @return array
@@ -209,7 +187,7 @@ class UpdateMediaItemForVideoType extends FormBase {
       }
     }
     catch (\Exception $e) {
-      $this->watchdogException($e);
+      $this->logger->watchdogException($e, 'Could not load mpx video type options.');
     }
 
     return $video_opts;
--- a/src/Form/UpdateMediaItemForVideoType.php
+++ b/src/Form/UpdateMediaItemForVideoType.php
@@ -123,7 +123,7 @@ class UpdateMediaItemForVideoType extends FormBase {
       // Up until here, all necessary checks have been made. No custom exception
       // handling needed other than for the db possibly exploding at this point.
       $this->messenger()->addError($this->t('There has been an unexpected problem updating the video. Check the logs for details.'));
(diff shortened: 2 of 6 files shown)


src/FormAlter/MediaFormAlter.php
```diff
@@ -0,0 +1,136 @@
...
+    ];
+  }
+
+  /**
+   * Callback for the 'reimport' button.
+   */
+  public function reimportCallback(array $form, FormStateInterface $formState) {
+    /* @var \Drupal\Core\Entity\ContentEntityForm $form_object */
+    $form_object = $formState->getFormObject();
+    $video = $form_object->getEntity();
```
salvamomo, line 77: @edutrul @juampynr Turns out I wasn't familiar (or had forgotten) that $form_state has a `getFormObject` method, and you can retrieve the entity from there, check if it's a `ContentEntityForm`, etc.  That solves the issue with `readonly` fields, as there's direct access available to the entity (without relying on the path).

Having said that, this is the biggest DrupalTF I've seen in a long time. Having a `$form` variable in form alters, which does **not** hold the form object, but instead it's $form_state the one that gives you a form object. Come on...

crying

salvamomo:
This is ready. I've skipped the automated test for this. Doing it via Kernel is complaining about 'html_date' being missing, when I try to build the form via the form builder. This leads to a broken test when `TimestampDatetimeWidget::formElement()` tried to render the widget.

While I'm sure it's probably due to some missing config / module to install, I don't think it's worth debugging that for the value that test adds at the moment (which is basically verifying that the "reimport" button, does trigger the `update` service. That service is already tested in isolation, so this UI part is not that important in itself.

media_mpx.module
```diff
@@ -179,3 +180,11 @@ function media_mpx_migration_plugins_alter(array &$migrations) {
     }
   }
 }
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function media_mpx_form_media_form_alter(&$form, $form_state, $form_id) {
+  $media_form_alter = MediaFormAlter::create(\Drupal::getContainer());
```
m4olivei: Why create the object this way vs. declaring it as a service and then doing `\Drupal::service('media_form_service')`?

src/FormAlter/MediaFormAlter.php
```diff
@@ -0,0 +1,148 @@
+<?php
+
+declare(strict_types = 1);
```
m4olivei, line 3: Interesting, this is the first time I've seen this. TIL. What led you to add this here? Is this something your adding to all new classes these days?

src/FormAlter/MediaFormAlter.php
```diff
@@ -0,0 +1,148 @@
...
+    if (!$form_object instanceof ContentEntityForm) {
+      return FALSE;
+    }
+
+    try {
+      $mpx_types = $this->mpxMediaTypeRepository->findAllTypes();
+      $type_ids = [];
+      foreach ($mpx_types as $key => $type) {
+        $type_ids[] = $type->id();
+        return in_array($form_object->getEntity()->bundle(), $type_ids);
```
m4olivei: Should this return statement be outside the `foreach` loop so that you have all the `$type_ids` in hand first before returning?

src/FormAlter/MediaFormAlter.php
```diff
@@ -0,0 +1,148 @@
...
+    $form_object = $formState->getFormObject();
+    $video = $form_object->getEntity();
+
+    $id_field = NULL;
+    if ($media_type = $this->mpxMediaTypeRepository->findByTypeId($video->bundle())) {
+      $field_map = $media_type->getFieldMap();
+      $id_field = $field_map['Media:id'] ?: NULL;
+    }
+
+    $mpx_id = $video->{$id_field}->value;
```
m4olivei: I'd suggest being more defensive here. `$id_field` is user input. It may or may not be defined, the field may or may not be empty, and the field may or may not have a `value` property. Maybe something like:

```
try {
  if (($field = $video->get($id_field)) && !$field->isEmpty()) {
    $mpx_id = $field->getString();
  }
}
catch (\InvalidArgumentException $e) {
}
```

src/FormAlter/MediaFormAlter.php
```diff
@@ -0,0 +1,148 @@
...
+  /**
+   * Update video data and show success / error messages as relevant.
+   */
+  public function updateVideoData(int $mpxId, string $mediaTypeId) {
+    $request = new UpdateVideoItemRequest($mpxId, $mediaTypeId);
+    try {
+      $this->updateService->execute($request);
+      $this->messenger()->addMessage($this->t('The video has been reimported.'));
+    }
+    catch (\Exception $e) {
```
m4olivei, line 117: Is there a more specific exception type to capture and report?

m4olivei (changes requested):

src/FormAlter/MediaFormAlter.php
salvamomo: @m4olivei I've been adding it whenever I remembered for the last few months. Just last week I actually added to my file templates in phpStorm. 

whenever that's declared in a file, it'll keep php from doing the automatic casting it usually does when, for example, you declare a typehint on a method argument, but then the user pass another type. By default, php would cast it. With this, it'll throw an error, so I'm basically enforcing that whoever is calling methods, is passing the correct data from the calling place, and doing there whatever casting is needed, explicitly.

src/FormAlter/MediaFormAlter.php
salvamomo: oh, I clearly overlooked this 👍

media_mpx.module
salvamomo: Mainly to keep the services file "clean" of services that in reality do nothing out of this context, given they still need to be called from this `_alter()`. Having said that, I've no real preference here, just thought it wasn't adding any more value for a case like :).

src/FormAlter/MediaFormAlter.php
salvamomo: @m4olivei No. The only possible ones are certainly well prevented before this point. This is just a catch all in case something completely unrelated breaks, to avoid a WSOD.

media_mpx.module
m4olivei: Fair enough. I like your rationale here, make a lot of sense. Especially as busy as the services file of this module already is!

src/FormAlter/MediaFormAlter.php
salvamomo: Great call. In fact I'm kind of surprised $id_field can pass at null. I'm pretty sure in an earlier version of the class, I was returning early, too 🤷‍♂ . I took this chance to improve the scenario where an user hasn't configured the media source at all (e.g: items being present already from a content import from D7), and added explicit messaging about that:

media_mpx_configure_media_source

salvamomo:
@m4olivei Thanks for the feedback! I think everything's addressed now, plus took the chance to make a small UX improvement. New screenshot attached in description 👍

m4olivei (approved):
Static review looks good. I admit to not having taken this for a spin yet.

juampynr:
@salvamomo can you fix conflicts?

juampynr:
@salvamomo thansk for the screenshots at the top. They really help.

salvamomo:
@juampynr I didn't see any conflicts, although it seems this project is configured to complain whenever the merged branch doesn't match 8.x-1.x latest. I just brought it up to date 👍

src/FormAlter/MediaFormAlter.php
```diff
@@ -0,0 +1,179 @@
...
+  /**
+   * Callback for the 'reimport' button.
+   */
+  public function reimportCallback(array $form, FormStateInterface $formState) {
+    /* @var \Drupal\Core\Entity\ContentEntityForm $form_object */
+    $form_object = $formState->getFormObject();
+    $video = $form_object->getEntity();
+
+    if (!$id_field = $this->resolveIdFieldName($video)) {
+      $this->messenger()->addError($this->t('The mpx field @field is not mapped to any local field. The Media source must be configured
```
deviantintegral: I'm not sure this is possible. When the media type is created, Drupal automatically creates a "source" field. See `\Drupal\media\MediaSourceBase::createSourceField()`.

If it is somehow possible, lets hide the button in the form if a source field isn't configured.

media_mpx.module
deviantintegral: I think a more common pattern would be to call the constructor here. Is this class ever actually used by core in a way where it calls create(), or is this the only call?

src/Form/UpdateMediaItemForVideoType.php
```diff
@@ -125,7 +123,7 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
       // Up until here, all necessary checks have been made. No custom exception
       // handling needed other than for the db possibly exploding at this point.
       $this->messenger()->addError($this->t('There has been an unexpected problem updating the video. Check the logs for details.'));
-      $this->watchdogException($e);
+      $this->logger->watchdogException($e, sprintf('mpx video with guid %s could not be updated', $guid));
```
deviantintegral: I think this should be message and context variables without the sprintf, so the text can translated if needed?

src/FormAlter/MediaFormAlter.php
```diff
@@ -0,0 +1,179 @@
...
+   * Implements hook_form_FORM_ID_alter().
+   */
+  public function alter(&$form, FormStateInterface $form_state, $form_id): void {
+    if ($this->alterAppliesToForm($form_state) === FALSE) {
+      return;
+    }
+
+    $form['actions']['reimport'] = [
+      '#type' => 'submit',
+      '#value' => $this->t('Reimport mpx data'),
```
deviantintegral: To me, "Update" sounds like more familiar language than "Reimport". Thoughts?

src/FormAlter/MediaFormAlter.php
```diff
@@ -0,0 +1,179 @@
...
+  /**
+   * Returns the name of the entity field holding the mpx ID.
+   *
+   * @param \Drupal\media\Entity\Media $video
+   *   The media entity for which to fetch the mpx ID field.
+   *
+   * @return string|null
+   *   The name of the field holding the mpx ID, or NULL if not configured.
+   */
+  private function resolveIdFieldName(Media $video):? string {
```
deviantintegral, line 100: I think instead call `\Drupal\media\MediaSourceInterface::getSourceFieldDefinition()` on the media entity.

src/FormAlter/MediaFormAlter.php
deviantintegral: Perfect. Normally this is bad, except this is UI code. 💯

src/FormAlter/MediaFormAlter.php
```diff
@@ -0,0 +1,179 @@
...
+   */
+  public function updateVideoData(int $mpxId, string $mediaTypeId) {
+    $request = new UpdateVideoItemRequest($mpxId, $mediaTypeId);
+    try {
+      $this->updateService->execute($request);
+      $this->messenger()->addMessage($this->t('The video has been reimported.'));
+    }
+    catch (\Exception $e) {
+      $this->messenger()->addError("The video data could not be reimported. Try again in a few minutes.");
+      $this->logger->watchdogException($e, sprintf('Video Reimport failed for item %s', $mpxId));
```
deviantintegral: message / context here again.

src/FormAlter/MediaFormAlter.php
```diff
@@ -0,0 +1,179 @@
...
+      return FALSE;
+    }
+
+    try {
+      $mpx_types = $this->mpxMediaTypeRepository->findAllTypes();
+      $type_ids = [];
+      foreach ($mpx_types as $key => $type) {
+        $type_ids[] = $type->id();
+      }
+      return in_array($form_object->getEntity()->bundle(), $type_ids);
```
deviantintegral: I think this method can be simplified to something like:

```
$form_object->getEntity()->getSource() instanceof \Drupal\media_mpx\Plugin\media\Source\MpxMediaSourceInterface
```

src/MpxLogger.php
```diff
@@ -84,7 +84,7 @@ public function logException(TransferException $exception, $severity = RfcLogLev
    *
    * @see \Drupal\Core\Utility\Error::decodeException()
    */
-  private function watchdogException(\Exception $exception, $message = NULL, array $variables = [], $severity = RfcLogLevel::ERROR, $link = NULL) {
+  public function watchdogException(\Exception $exception, $message = NULL, array $variables = [], $severity = RfcLogLevel::ERROR, $link = NULL) {
```
deviantintegral, line 87: can we get this in core yet 😂

deviantintegral (commented):
Left some notes. This is close!

src/FormAlter/MediaFormAlter.php
salvamomo: @deviantintegral covering for cases where the data is coming from a migration into an existing type. Field mappings might not be applied yet, and this can effectively return null. 

I'll hide the button in that case 👍

src/MpxLogger.php
salvamomo: yeah... 😭

src/FormAlter/MediaFormAlter.php
deviantintegral: @juampynr seems like this could be a good improvement for the migration - it should make defining a source field required I think.

src/Form/UpdateMediaItemForVideoType.php
salvamomo: 🤦‍♂  I misread the usage of the watchdogException method, and thought it wasn't passing the $variables for that. Changing.

src/FormAlter/MediaFormAlter.php
salvamomo: Not sure if I'm missing something as I'm slightly new to Media in D8, but I'm not able to do that. It still requires me to pass the MediaType, which I need to load for that, so once loaded, it seems more direct to just call FieldMap in it and make the check.

salvamomo:
@deviantintegral Everything is tackled now I think 👍

src/FormAlter/MediaFormAlter.php
deviantintegral: You can call `$media->bundle->entity` to get the media type - example at https://github.com/nbc-bravo/core/blob/fd294077e54a94275201474c3d7b3979c874b671/modules/media_library/src/Form/MediaLibraryUploadForm.php#L141-L143

salvamomo:
@deviantintegral Thanks for the pointer to https://github.com/nbc-bravo/core/blob/fd294077e54a94275201474c3d7b3979c874b671/modules/media_library/src/Form/MediaLibraryUploadForm.php#L141-L143. 

I've updated the `resolveFieldId` method to make use of that to retrieve the field map. I can't see `getSourceFieldDefinition` gives me what I need and the usage seems a bit more convoluted, so I've stuck to `->getFieldMap()`, which seems to be used in other places too. Nevertheless, this allows to get rid of the repository dependency 👍 .

PS: This Drupal way of accessing these items via properties with magic methods makes me itch a bit. And Media module itself does it that way! :(.

salvamomo:
Updated from 8.x-1.x. Merging after tests pass.

PR #131, opened 2019-05-29
UI: Removed traces of field_mpx_id field mapping

salvamomo:
## Summary of changes.

- Removes 1 remaining usage of `field_mpx_id`.
- Simplifies a couple forms and allows to use the update alter one even if the **Media:Id** mapping is not configured.
- Fixes some PHPCS errors reported by Circle. They're not related to this PR, and the code that introduced them is several months old, so I'm not sure of why Circle didn't complain before!

### DIFF
--- a/src/FormAlter/MediaFormAlter.php
+++ b/src/FormAlter/MediaFormAlter.php
@@ -74,15 +74,9 @@ class MediaFormAlter {
   public function reimportCallback(array $form, FormStateInterface $formState) {
     /* @var \Drupal\Core\Entity\ContentEntityForm $form_object */
     $form_object = $formState->getFormObject();
-    $video = $form_object->getEntity();
-
-    $id_field = $this->resolveIdFieldName($video);
 
     try {
-      if (($field = $video->get($id_field)) && !$field->isEmpty()) {
-        $mpx_id = $video->{$id_field}->value;
-        $this->updateVideoData((int) $mpx_id, $video->bundle());
-      }
+      $this->updateVideoData($form_object->getEntity());
     }
     catch (\InvalidArgumentException $e) {
     }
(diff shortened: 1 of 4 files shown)


src/FormAlter/MediaFormAlter.php
```diff
@@ -98,25 +92,26 @@ public function reimportCallback(array $form, FormStateInterface $formState) {
    *   The name of the field holding the mpx ID, or NULL if not configured.
    */
   private function resolveIdFieldName(Media $video):? string {
-    if (!$field_map = $video->bundle->entity->getFieldMap()) {
-      return NULL;
-    }
-
-    return isset($field_map['Media:id']) ? $field_map['Media:id'] : NULL;
+    $media_source = $video->getSource();
```
salvamomo, line 95: Turns out this field mapping is not mandatory (which was an assumption I'd made when developing all the UI work). The value from the source field definition is the reliable source value we need.

src/FormAlter/MediaFormAlter.php
```diff
@@ -98,25 +92,26 @@ public function reimportCallback(array $form, FormStateInterface $formState) {
...
-    $request = new UpdateVideoItemRequest($mpxId, $mediaTypeId);
+  public function updateVideoData(Media $video) {
+    $request = UpdateVideoItemRequest::createFromMediaEntity($video);
     try {
       $this->updateService->execute($request);
       $this->messenger()->addMessage($this->t('The video has been reimported.'));
     }
     catch (\Exception $e) {
+      $id_field = $this->resolveIdFieldName($video);
+      $mpx_id = $video->{$id_field}->value;
```
salvamomo, line 111: This can't really be empty, so I've removed the isEmpty() checks.

src/Service/UpdateVideoItem/UpdateVideoItemRequest.php
```diff
@@ -51,9 +51,12 @@ public function __construct(int $mpxId, string $mediaTypeId) {
    *   The request object, to be passed to the UpdateVideoItem service.
    */
   public static function createFromMediaEntity(Media $entity): UpdateVideoItemRequest {
-    $mpx_id = (int) $entity->get('field_mpx_id')->value;
```
salvamomo, line 54: This `field_mpx_id` reference had made it through. Just noticed yesterday.

src/Service/UpdateVideoItem/UpdateVideoItemRequest.php
```diff
@@ -51,9 +51,12 @@ public function __construct(int $mpxId, string $mediaTypeId) {
...
-    $media_type_id = $entity->bundle();
-    return new self($mpx_id, $media_type_id);
+    $media_source = $entity->bundle->entity->getSource();
+    $source_field = $media_source->getSourceFieldDefinition($entity->bundle->entity);
+
+    $global_url_parts = explode('/', $entity->get($source_field->getName())->value);
+    $mpx_id = (int) end($global_url_parts);
```
deviantintegral, line 58: Now that I see this, we are trying to be consistent and use the full IDs. The API makes no guarantees that just the numeric part of the ID is unique, though I think in practice it is. I'd say this class should construct either with a string or with a Uri object like `\Lullabot\Mpx\DataService\DataObjectFactory::load`.

salvamomo: @deviantintegral Originally I started passing the Uri object, but then you pointed out that wasn't needed as there's a method to do it with the numeric one in the DataObjectImporter (which is handy, just a shame we can't rely on that from mpx :( ). Since other parts are already relying on this, and in any case this PR does some improvements and a bugfix, can we open an issue for that? That way it stays documented, even if not tackled immediately.

deviantintegral: Yeah, in fact I have a few followups to file from the demo. Thanks!

PR #132, opened 2019-05-29
Feature: Delete http method to unpublish entities

germanmontesm:
delete http method is now handled on the notification queue worker, added method to unpublish related entities

### DIFF
--- a/src/DataObjectImporter.php
+++ b/src/DataObjectImporter.php
@@ -89,15 +89,19 @@ class DataObjectImporter {
    *   The array of media entities that were imported.
    */
   public function unpublishItem(int $mpx_object_id, MediaTypeInterface $media_type): array {
+    // Find any existing media items, or return a new one.
     $results = $this->loadMediaEntitiesById($media_type, $mpx_object_id);
     if (!empty($results)) {
       foreach ($results as $entity) {
+        // Check if current entity is published or not
         if ($entity->isPublished()) {
+          // If yes then set as unpublished and save this change
           $entity->setUnpublished();
           $entity->save();
         }
       }
     }
+    // Return unpublished entities
     return $results;
   }
 
@@ -118,7 +122,6 @@ class DataObjectImporter {
     // static cache.
     $reset_ids = [];
     // @todo start a transaction.
-
     // Find any existing media items, or return a new one.
     $results = $this->loadMediaEntities($media_type, $mpx_object);
 
--- a/src/DataObjectImporter.php
+++ b/src/DataObjectImporter.php
@@ -90,7 +90,8 @@ class DataObjectImporter {
    */
   public function unpublishItem(int $id, MediaTypeInterface $media_type): array {
     // Find any existing media items.
-    foreach ($this->loadMediaEntitiesById($media_type, $id) as $entity) {
+    $entities = $this->loadMediaEntitiesById($media_type, $id);
+    foreach ($entities as $entity) {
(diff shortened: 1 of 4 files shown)


src/DataObjectImporter.php
```diff
@@ -146,6 +166,26 @@ private function loadMediaEntities(MediaTypeInterface $media_type, ObjectInterfa
...
+   *   The mpx object id to load the associated entities for.
+   *
+   * @return \Drupal\media\Entity\Media[]
+   *   An array of existing media entities or a new media entity.
+   */
+  private function loadMediaEntitiesById(MediaTypeInterface $media_type, int $mpx_object_id): array {
+    $media_source = $this->loadMediaSource($media_type);
+    $source_field = $media_source->getSourceFieldDefinition($media_type);
+    $media_storage = $this->entityTypeManager->getStorage('media');
+    $results = $media_storage->loadByProperties([$source_field->getName() => (string) $mpx_object_id]);
```
deviantintegral: I'm not sure this works.

mpx "full IDs" are URIs ending with an integer, such as this example: https://github.com/Lullabot/media_mpx/blob/8.x-1.x/tests/fixtures/media-object.json#L5

So, if we're passing in just the integer component (from the `int` typehint), casting that to a string still won't have the full URI, and $results will always be empty.

Throughout the rest of the code we try to just keep the full URI as a string or Guzzle URI object - can we do that here?

src/Plugin/QueueWorker/NotificationQueueWorker.php
```diff
@@ -154,10 +154,17 @@ private function yieldLoads(array $notifications): \Generator {
     foreach ($notifications as $notification) {
       /** @var \Lullabot\Mpx\DataService\Media\Media $mpx_media */
       $mpx_media = $notification->getNotification()->getEntry();
+      switch ($notification->getNotification()->getMethod()) {
+        case "delete":
+          $this->importer->unpublishItem($notification->getNotification()->getId(), $notification->getMediaType());
```
deviantintegral: I think this will be the id of the notification object, not the id of the media object. Does this need to be:

`$notification->getNotification()->getEntry()->getId()`?

deviantintegral (commented):
A few questions, but this is close!

There's a todo we can remove at https://github.com/Lullabot/media_mpx/blob/73c23f0cde0ea11a567f86ea8de3060ebdc36874/src/DataObjectImporter.php#L114-L115

Go ahead and fix the two typehinting changes, as it looks like coder module has updated it's rules.

Note the test failures at https://circleci.com/gh/Lullabot/media_mpx/2546?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link - I think they confirm what I'm seeing on a static review.

src/DataObjectImporter.php
germanmontesm: done.

src/Plugin/QueueWorker/NotificationQueueWorker.php
germanmontesm: done.

deviantintegral:
Static review looks good, I'd like to test it before merging.

deviantintegral (approved):
To test this I modified `vendor/lullabot/mpx-php/src/DataService/Notification.php`'s `getMethod` method to return `delete`, which comes as close as we can get to testing without actually deleting videos. I then ran notifications and saw the video become unpublished. Nice work!

PR #133, opened 2019-05-31
Clean up text for video import forms

deviantintegral:


juampynr:
Looks good to me!

salvamomo (approved):
Great changes! I was a bit hesitant about the queue VS import terminology, as it could lead editors to think the actual video import was happening in that exact moment. We can always add more description text in the queue page if that ever happens 👍

PR #140, opened 2019-07-03
GH-139: Added support for timestamp dates on MediaAvailableAccess::getDateTime()

### LINKED ISSUE #139: MediaAvailableAccess::getDateTime() returns null date when the mpx object date field contains a timestamp
### Motivation
I'm finding a problem where $entity->access('view') is being called for a video entity that is past its expiration date, yet it's returning access allowed results.

Digging into it, what happens is that `MediaAvailableAccess::getDateTime()` is returning NULL results event though there's an actual value in the `MediaInterface $media`. The value is a valid timestamp, whereas this method only has support for a specific format.

(https://github.com/Lullabot/media_mpx/issues/134) https://github.com/Lullabot/media_mpx/pull/135 the symptom, which was a php error, but the problem still needs fixing; it just became silent.

### Solution
Expand the function to make it cope with timestamp values as well.

salvamomo:
- Fixes https://github.com/Lullabot/media_mpx/issues/139
- Fixes some PHPCS errors that appeared, though they were unrelated to this PR.
- Adds kernel test to cover availability window access checks on `MediaAvailableAccess` service.

- Opened https://github.com/Lullabot/media_mpx/issues/141.

### DIFF
--- a/src/Access/MediaAvailableAccess.php
+++ b/src/Access/MediaAvailableAccess.php
@@ -206,9 +206,36 @@ class MediaAvailableAccess {
    */
   private function getDateTime(MediaInterface $media, string $field_name): DateTimeFormatInterface {
     $fieldItemList = $media->get($field_name);
-    $date_time = \DateTime::createFromFormat(DateTimeItemInterface::DATETIME_STORAGE_FORMAT, $fieldItemList->value);
-    $date = !$fieldItemList->isEmpty() && $date_time ? new ConcreteDateTime($date_time) : new NullDateTime();
-    return $date;
+    if ($fieldItemList->isEmpty()) {
+      return new NullDateTime();
+    }
+
+    if ($date_time = \DateTime::createFromFormat(DateTimeItemInterface::DATETIME_STORAGE_FORMAT, $fieldItemList->value)) {
+      return new ConcreteDateTime($date_time);
+    }
+
+    // One last attempt to get a date time object if the value is a timestamp.
+    if ($this->isValidTimestamp($fieldItemList->value)) {
+      $date_time = (new \DateTime())->setTimestamp($fieldItemList->value);
+      return new ConcreteDateTime($date_time);
+    }
+
+    return new NullDateTime();
+  }
+
+  /**
+   * Checks whether a value is valid to be used as a timestamp.
+   *
+   * @param string|int $value
+   *   The value to check.
+   *
+   * @return bool
+   *   TRUE if the value is a valid timestamp, FALSE otherwise.
+   */
+  private function isValidTimestamp($value) {
+    return ((string) (int) $value === $value)
+      && ($value <= PHP_INT_MAX)
+      && ($value >= ~PHP_INT_MAX);
   }
 
 }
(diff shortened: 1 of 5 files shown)


src/Access/MediaAvailableAccess.php
```diff
@@ -206,9 +206,36 @@ protected function getStubMediaObject(MediaInterface $media): MpxMedia {
...
+   * @param string|int $value
+   *   The value to check.
+   *
+   * @return bool
+   *   TRUE if the value is a valid timestamp, FALSE otherwise.
+   */
+  private function isValidTimestamp($value) {
+    return ((string) (int) $value === $value)
+      && ($value <= PHP_INT_MAX)
+      && ($value >= ~PHP_INT_MAX);
```
deviantintegral: It's been a long time since I've seen bitwise operators in PHP code. Neat! Is there a reason you're not using `PHP_INT_MIN`?

This isn't a valid timestamp, at least according to a typical 32-bit unix timestamp field. Should we constrain this to 32 bits?

deviantintegral (changes requested):
I think there's actually a different bug at play here:

https://github.com/Lullabot/media_mpx/blame/fa92229b30a1079d81a742a0af270cd2a6b02279/src/Plugin/media/Source/MediaSourceBase.php#L265-L269

That section wasn't intended to be added to this repository (oops past Andrew!). In summary, during migration we missed converting int fields to date / time fields (which are better because they are explicit about time zones). As a workaround, we patched in formatting the dates as a timestamp.

At this point, we could either go all-in and support timestamp fields everywhere (though we should also support mapping to date time fields), or roll that back and update the migration to convert the field types and data.

deviantintegral:
Oh, and whatever we do we should add tests because clearly this is a pain point.

deviantintegral:
The CS failures are likely because coder module added additional checks. We should either fix them or disable the new sniffs, which would be a good straight-forward PR to get done before this one.

src/Access/MediaAvailableAccess.php
salvamomo: I've seen in other cases php_int_min defined as the opposite of max, but no real reason. I've actually removed this method entirely. For the purposes of this method, `createFromFormat('U')` is actually a safer and simpler choice.

salvamomo:
Tests are ready and phpcs warnings cleaned up. This is ready 👍

deviantintegral (approved):

deviantintegral:
Good stuff! Thanks.

m4olivei:
This is fantastic! The bravo fork had a hack around this issue. Wonderful to see it addressed more formally.

tests/src/Kernel/Access/MediaAvailableAccessTest.php
```diff
@@ -0,0 +1,256 @@
...
+      $access = $available_access_service->view($media, $anonymous);
+      $this->assertInstanceOf($expectedClass, $access);
+    }
+  }
+
+  /**
+   * Data provider of available and expiration dates as timestamps.
+   *
+   * Not used via the @dataProvider annotation on purpose, as that'll make each
+   * iteration run the setUp logic, which is not desired for these tests.
```
juampynr, line 120: :+1:

PR #143, opened 2019-07-05
Issue #142 Import single messaging for filtered videos

### LINKED ISSUE #142: Import mpx video by GUID can report false positives
While testing the Import mpx video by GUID UI, I ran into a situation where I attempted to import a single video, the UI reported success with this message:

> The selected video has been imported.

However, the video was no where to be found. As it turns out, the guid that I was using was correctly being filtered out by a custom import event subscriber.

[From a glance at the code](https://github.com/Lullabot/media_mpx/blob/8.x-1.x/src/Service/UpdateVideoItem/UpdateVideoItem.php#L72), it appears that the submit handler for the Import mpx video by GUID UI could detect this case and report back something more informative. As it is, the message is sort of a false positive. Yes, the command was successful, but it didn't actually import anything. It looks like this would be the same problem on the Import mpx video by Full ID UI as well.

m4olivei:
Resolves #142 using a generic warning message to indicate to the user that something didn't import. Better would be if we could be more specific. That would take some additional work, so I felt it was worth the small increment first.

Also tidies up the copy on all the import forms a bit to be more consistent.

### DIFF
--- a/src/Form/ImportVideoItemByMpxId.php
+++ b/src/Form/ImportVideoItemByMpxId.php
@@ -79,8 +79,8 @@ class ImportVideoItemByMpxId extends FormBase {
     ];
     $form['mpx_id'] = [
       '#type' => 'textfield',
-      '#title' => t('ID'),
-      '#placeholder' => 'Type the ID of the mpx video you want to import.',
+      '#title' => $this->t('ID'),
+      '#placeholder' => $this->t('Type the ID of the mpx video you want to import.'),
       '#required' => FALSE,
     ];
     $form['actions']['#type'] = 'actions';
--- a/src/Form/ImportVideoItemByMpxId.php
+++ b/src/Form/ImportVideoItemByMpxId.php
@@ -7,6 +7,7 @@ namespace Drupal\media_mpx\Form;
 use Drupal\Core\Form\FormBase;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\media_mpx\MpxLogger;
+use Drupal\media_mpx\Plugin\media\Source\Media;
 use Drupal\media_mpx\Repository\MpxMediaType;
 use Drupal\media_mpx\Service\UpdateVideoItem\UpdateVideoItem;
 use Drupal\media_mpx\Service\UpdateVideoItem\UpdateVideoItemRequest;
@@ -104,9 +105,9 @@ class ImportVideoItemByMpxId extends FormBase {
       $response = $this->updateVideoItem->execute($request);
       if (empty($response->getUpdatedEntities())) {
         $mpx_media = $response->getMpxItem();
-        $this->messenger()->addWarning($this->t("The selected video: @video_title (@guid) did not import. The video was filtered out by one or more custom import filters. Adjust the video metadata in mpx to ensure it's available to be imported and try again.", [
+        $this->messenger()->addWarning($this->t("The selected video: @video_title (@id) did not import. The video was filtered out by one or more custom import filters. Adjust the video metadata in mpx to ensure it's available to be imported and try again.", [
           '@video_title' => $mpx_media->getTitle(),
-          '@guid' => $mpx_media->getGuid(),
+          '@id' => Media::getMpxObjectIdFromUri((string) $mpx_media->getId()),
         ]));
       }
       else {
@@ -116,7 +117,7 @@ class ImportVideoItemByMpxId extends FormBase {
     catch (\Exception $e) {
       // Up until here, all necessary checks have been made. No custom exception
       // handling needed other than for the db possibly exploding at this point.
-      $this->messenger()->addError($this->t('There has been an unexpected problem updating the video. Check the logs for details.'));
+      $this->messenger()->addError($this->t('There has been an unexpected problem importing the video. Check the logs for details.'));
       $this->logger->watchdogException($e);
     }
   }
--- a/src/Form/ImportVideoItemByMpxId.php
+++ b/src/Form/ImportVideoItemByMpxId.php
@@ -98,8 +98,16 @@ class ImportVideoItemByMpxId extends FormBase {
 
     $request = new UpdateVideoItemRequest($mpx_id, $video_type);
     try {
-      $this->updateVideoItem->execute($request);
-      $this->messenger()->addMessage($this->t('The selected video has been imported.'));
+      $response = $this->updateVideoItem->execute($request);
+      if (empty($response->getUpdatedEntities())) {
+        $this->messenger()->addWarning($this->t('The selected video: @video_title (@guid) did not import. The video was filtered out by one or more custom import filters. Adjust the video metadata in mpx to ensure it\'s available to be imported and try again.', [
+          '@video_title' => $mpx_media->getTitle(),
+          '@guid' => $mpx_media->getGuid(),
+        ]));
+      }
+      else {
+        $this->messenger()->addMessage($this->t('The selected video has been imported.'));
+      }
     }
     catch (\Exception $e) {
       // Up until here, all necessary checks have been made. No custom exception
--- a/src/Form/ImportVideoItemByMpxId.php
+++ b/src/Form/ImportVideoItemByMpxId.php
@@ -71,19 +71,22 @@ class ImportVideoItemByMpxId extends FormBase {
 
     $form['video_type'] = [
       '#type' => 'select',
-      '#title' => $this->t('Video Type'),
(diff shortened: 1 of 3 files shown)


src/Form/QueueContentsForm.php
```diff
@@ -50,13 +50,15 @@ public function buildForm(array $form, FormStateInterface $form_state) {
 
     $form['mpx_video_types'] = [
       '#type' => 'checkboxes',
-      '#title' => $this->t('mpx Video Types'),
+      '#title' => $this->t('Mpx video types'),
```
salvamomo: Certainly odd at first sight, but I left this as lowercase `mpx` on purpose, as per https://github.com/Lullabot/media_mpx#about-the-mpx-name. 

Fine with the change if I got something wrong, 👍

src/Form/UpdateMediaItemForVideoType.php
```diff
@@ -153,9 +154,8 @@ private function submitFormProcessRequest(UpdateVideoItemRequest $request, strin
       $this->submitFormExecuteRequest($request, $success_text);
     }
     catch (\Exception $e) {
-      // Up until here, all necessary checks have been made. No custom
-      // exception handling needed other than for the db possibly
-      // exploding at this point.
+      // Up until here, all necessary checks have been made. No custom exception
+      // handling needed other than for the db possibly exploding at this point.
```
salvamomo, line 158: 💯 Thanks for putting this back into the original line length, as per CS!

salvamomo (commented):
This looks good to me, just a comment on the usage of the "mpx" name. Otherwise good to go 🚀

src/Form/ImportVideoItemByMpxId.php
```diff
@@ -71,19 +72,22 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+      '#title' => $this->t('Video type'),
       '#description' => $this->t('Choose the video type to import the video into.'),
       '#options' => $video_opts,
       '#required' => TRUE,
     ];
     $form['mpx_id'] = [
       '#type' => 'textfield',
-      '#title' => t('mpx item ID'),
+      '#title' => t('ID'),
+      '#placeholder' => 'Type the ID of the mpx video you want to import.',
```
juampynr: No `t()` ?

src/Form/ImportVideoItemByMpxId.php
```diff
@@ -98,13 +102,22 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
 
     $request = new UpdateVideoItemRequest($mpx_id, $video_type);
     try {
-      $this->updateVideoItem->execute($request);
-      $this->messenger()->addMessage($this->t('The selected video has been imported.'));
+      $response = $this->updateVideoItem->execute($request);
+      if (empty($response->getUpdatedEntities())) {
+        $mpx_media = $response->getMpxItem();
+        $this->messenger()->addWarning($this->t("The selected video: @video_title (@id) did not import. The video was filtered out by one or more custom import filters. Adjust the video metadata in mpx to ensure it's available to be imported and try again.", [
```
juampynr: Would this message be shown on screen? If so, I wonder if it will be correct for an editor.

src/Form/UpdateMediaItemForVideoType.php
```diff
@@ -169,8 +169,17 @@ private function submitFormProcessRequest(UpdateVideoItemRequest $request, strin
    *   Success text to be used on the notification.
    */
   private function submitFormExecuteRequest(UpdateVideoItemRequest $request, string $success_text) {
-    $this->updateVideoItemService->execute($request);
-    $this->messenger()->addMessage($success_text);
+    $response = $this->updateVideoItemService->execute($request);
+    if (empty($response->getUpdatedEntities())) {
+      $mpx_media = $response->getMpxItem();
+      $this->messenger()->addWarning($this->t("The selected video: @video_title (@guid) did not import. The video was filtered out by one or more custom import filters. Adjust the video metadata in mpx to ensure it's available to be imported and try again.", [
```
juampynr: Same as above. If this will show up on screen, I wonder if an editor would know what an import filter is.

src/Form/ImportVideoItemByMpxId.php
m4olivei: Good catch, will fix.

src/Form/ImportVideoItemByMpxId.php
m4olivei: Yes, it will be shown on screen. Editors definitely wouldn't know what an import filter is. 🤔 The most we can tell at this point is that the video was filtered out. We don't know why yet. There is some discussion about how we could eventually be more specific about why here: https://github.com/Lullabot/media_mpx/issues/142. Here is some alternate language off the top of my head, any of these sound better?

> The selected video: @video_title (@id) did not import. Adjust the video metadata in mpx to ensure it's available to be imported and try again.

> The selected video: @video_title (@id) did not import. One or more custom business rules are in place which filtered it out (for example Unapproved videos may be filtered out). Adjust the video metadata in mpx to ensure it's available to be imported and try again.

Shout out the site administrator:

> The selected video: @video_title (@id) did not import. There may be one or more custom business rules in place which filtered it out. Consult the site administrator and adjust the video metadata in mpx to ensure it's available to be imported and try again.

juampynr: I think that the latter would be the most clear to me for the short term. As for the long term, #142 would make this crystal clear.

src/Form/QueueContentsForm.php
m4olivei: Yeah, it is strange at second glance. I was following Drupal convention here which capitalizes the first word of a form label and lower cases the rest. This is a bit of an edge case, since the first word is an acronym that refuses to be all caps 😆 . Maybe the middle ground is `mpx Video types`. There is some precedence for this in the mpx documentation:

mpx_capitalization

src/Form/ImportVideoItemByMpxId.php
deviantintegral: Can we use StringTranslationTrait too?

m4olivei: Yep absolutely, it's already being used in this class.

src/Form/ImportVideoItemByMpxId.php
```diff
@@ -71,19 +72,22 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+      '#title' => t('ID'),
+      '#placeholder' => 'Type the ID of the mpx video you want to import.',
       '#required' => FALSE,
     ];
-    $form['submit'] = [
+    $form['actions']['#type'] = 'actions';
+    $form['actions']['submit'] = [
       '#type' => 'submit',
-      '#value' => $this->t('Import Item'),
+      '#value' => $this->t('Import video'),
```
deviantintegral, line 89: 👍

src/Form/UpdateMediaItemForVideoType.php
```diff
@@ -96,21 +96,22 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+      '#title' => $this->t('Video type'),
       '#description' => $this->t('Choose the video type to import the video into.'),
       '#options' => $video_opts,
       '#required' => TRUE,
     ];
     $form['guid'] = [
       '#type' => 'textfield',
-      '#title' => $this->t('guid'),
-      '#placeholder' => 'Type the GUID of the mpx item you want to import.',
+      '#title' => $this->t('Guid'),
```
deviantintegral: `GUID`?

src/Form/UpdateMediaItemForVideoType.php
```diff
@@ -131,8 +132,8 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
       $this->submitFormProcessRequest($request, $guid, $success_text, $error_text);
       return;
     }
-    $guid_not_found_exception = new \Exception("Given GUID doesn't exist, please check and try again.");
-    $error_text = (string) $this->t("Given GUID doesn't exist, please check and try again.");
+    $guid_not_found_exception = new \Exception("Given guid doesn't exist, please check and try again.");
```
deviantintegral: Can we `sprintf` in the GUID into this string (and the one below?).

Instead of `\Exception`, perhaps RuntimeException or DomainException?

deviantintegral (commented):
A few minor notes, thanks for this!

m4olivei:
Updates made! Thanks all for your suggestions.

PR #145, opened 2019-07-11
Issue #144 Fix an issue with serializing the media form for mpx entities

### LINKED ISSUE #144: LogicException: The database connection is not serializable error on mpx media forms with AJAX operations
We're getting the following error on any AJAX operations on our mpx media forms with the most recent version of 8.x-1.x:

```
LogicException: The database connection is not serializable. This probably means you are serializing an object that has an indirect reference to the database connection. Adjust your code so that is not necessary. Alternatively, look at DependencySerializationTrait as a temporary solution. in Drupal\Core\Database\Connection->__sleep() (line 1481 of /var/www/web/core/lib/Drupal/Core/Database/Connection.php).
```

I've so far narrowed it down to the new form alter here: https://github.com/Lullabot/media_mpx/blob/8.x-1.x/media_mpx.module#L187,L193

Which was this recent feature: https://github.com/Lullabot/media_mpx/pull/127

I further narrowed it down to one or more Connection objects in the `\Drupal\media_mpx\Service\UpdateVideoItem\UpdateVideoItem` members, which is a dependency of the `\Drupal\media_mpx\FormAlter\MediaFormAlter` class, which is included in the `$form` render array that is attempted to be cached during an AJAX operation.

We should be able to use `\Drupal\Core\DependencyInjection\DependencySerializationTrait` in  `\Drupal\media_mpx\Service\UpdateVideoItem\UpdateVideoItem` to resolve it. 

Will try to come up with a test to show the issue as well.

m4olivei:
Resolves #144.

### DIFF
--- a/src/Service/UpdateVideoItem/UpdateVideoItem.php
+++ b/src/Service/UpdateVideoItem/UpdateVideoItem.php
@@ -4,6 +4,7 @@ declare(strict_types = 1);
 
 namespace Drupal\media_mpx\Service\UpdateVideoItem;
 
+use Drupal\Core\DependencyInjection\DependencySerializationTrait;
 use Drupal\media_mpx\DataObjectFactoryCreator;
 use Drupal\media_mpx\DataObjectImporter;
 use Drupal\media_mpx\Repository\MpxMediaType;
@@ -14,6 +15,7 @@ use Drupal\media_mpx\Repository\MpxMediaType;
  * @package Drupal\media_mpx\Service\ImportVideoItem
  */
 class UpdateVideoItem {
+  use DependencySerializationTrait;
 
   /**
    * The Data Object Importer.


m4olivei:
Sadly, I don't think I will have time to write a test for this, as I'm out for the next few weeks and have some things to wrap up. Need the fix first.

I did some thinking about a test. The minimal reproducible test case would load the mpx media entity form, with any field widget that uses `#ajax`. In core, the only field that comes to mind is the file field. Otherwise we could do a custom widget just to add a silly `#ajax`. It triggers core to cache the form object and fail without this fix.

salvamomo (approved):
A test would be certainly ideal, but I think we can live without it, as this is an infrastructure dependency, unlikely to be removed as part of any logic changes anyway, so LGTM! 👍.

I'm still not convinced this should be serialized at all. Feels weird, though I get form API wants to do its thing 😅. It just occurred to me, another approach could be to declare the service lazy, as then what would be serialized would be the proxy class.  Nevertheless, this fixes the problem, so good to go! 💯

PR #146, opened 2019-07-18
Remove foreach and clean up form page titles

deviantintegral:
Instead of foreach, we can just pop the first video off of the iterator. This also cleans up the form page titles.

Screen Shot 2019-07-18 at 1 58 36 PM


juampynr:
Looks good to me.

salvamomo (approved):
Looking good! 👍

PR #148, opened 2019-07-21
Test against Drupal 8.7

deviantintegral:
Resolves all deprecation warnings and gets tests passing on Drupal 8.7.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,11 +16,11 @@
 defaults: &defaults
   docker:
     # specify the version you desire here (avoid latest except for testing)
-    - image: andrewberry/drupal_tests:0.4.0
+    - image: quay.io/deviantintegral/drupal_tests:latest
 
-    - image: selenium/standalone-chrome-debug:3.14.0-beryllium
+    - image: selenium/standalone-chrome-debug:3.141.59-neon
 
-    - image: mariadb:10.3
+    - image: mariadb:10.4
       environment:
         MYSQL_ALLOW_EMPTY_PASSWORD: 1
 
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -16,11 +16,11 @@
 defaults: &defaults
   docker:
     # specify the version you desire here (avoid latest except for testing)
-    - image: andrewberry/drupal_tests:0.4.0
+    - image: quay.io/deviantintegral/drupal_tests:latest
```
juampynr, line 19: :+1:

deviantintegral: Hah, I need to point this to a tag. Will follow up.

tests/src/FunctionalJavascript/FieldMapTest.php
```diff
@@ -39,6 +39,12 @@ class FieldMapTest extends WebDriverTestBase {
    *   - Validates saving an mpx media entity and that the field is visible.
    */
   public function testMapCustomField() {
+    $this->container->get('config.factory')
+      ->getEditable('media.settings')
+      ->set('standalone_url', TRUE)
+      ->save(TRUE);
+    $this->container->get('router.builder')->rebuild();
```
salvamomo, line 46: Curious, what was the thing that needed the rebuild here?

deviantintegral: The above media setting (FALSE by default for new installs) re-enables the canonical media view. Otherwise, the test fails because /media/1 redirects to /media/1/edit. This is copied from media module tests.

salvamomo: Cool, thanks for the explanation!

PR #150, opened 2019-07-25
Fix using the wrong string for update permissions

deviantintegral:
I really wish core used a constant or class so this could be a compile time and not runtime error that silently fails!


juampynr:
Looks good to me.

tests/src/Kernel/Access/MediaAvailableAccessTest.php
```diff
@@ -56,11 +58,7 @@ public function setUp() {
    */
   public function testAccessChecksForMediaItemWithTimestampFields() {
     $anonymous = User::getAnonymousUser();
-    $media = Media::create([
-      'bundle' => $this->mediaType->get('id'),
-      'title' => 'test available access',
-      'field_media_media_mpx_media' => 'http://example.com/1234',
-    ]);
+    $media = $this->createMedia();
```
salvamomo, line 61: 👌

Issue #151, opened 2019-07-27
If there is one element in dropdown in MPX admin UI then should be default one selected


edutrul:
The following field option values should be selected by default since there are no more option values:

/admin/content/media/mpx/queue-videos

 
![image](https://user-images.githubusercontent.com/1582129/61990295-4f19eb80-b003-11e9-8767-84796a07d571.png)

/admin/content/media/mpx/update-item-for-video-type

 
![image](https://user-images.githubusercontent.com/1582129/61990298-54773600-b003-11e9-8b6d-d7e5bc393fde.png)

/admin/content/media/mpx/import-item-by-mpx-id

 
![image](https://user-images.githubusercontent.com/1582129/61990300-59d48080-b003-11e9-8693-768fe978c295.png)

PR #152, opened 2019-07-27
#151: Default values should be selected

edutrul:
PR: https://github.com/Lullabot/media_mpx/pull/152

The following field option values should be selected by default since there are no more option values:

/admin/content/media/mpx/queue-videos

 
image

/admin/content/media/mpx/update-item-for-video-type

 
image

/admin/content/media/mpx/import-item-by-mpx-id

image


edutrul:
Screenshot as evidence:

image


image


image

edutrul:
PR up and running: https://github.com/Lullabot/media_mpx/pull/152

src/Form/ImportUpdateVideoItem.php
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    return $video_opts;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getFormId() {
+    return 'media_mpx_asset_sync_single_by_mpx_id';
```
salvamomo: This can't be in an abstract class, as different forms need different ids.

src/Form/ImportVideoItemByMpxId.php
```diff
@@ -65,31 +31,14 @@ public static function create(ContainerInterface $container) {
...
       '#placeholder' => $this->t('Type the ID of the mpx video you want to import.'),
       '#required' => FALSE,
     ];
-    $form['actions']['#type'] = 'actions';
-    $form['actions']['submit'] = [
-      '#type' => 'submit',
-      '#value' => $this->t('Import video'),
-      '#button_type' => 'primary',
-    ];
-
```
salvamomo, line 92: I don't think any of the `buildForm` contents should be moved into an abstract class in order to save a few a lines of code. I don't see this as being duplication that needs to be removed: it's 2 different forms, that do things with 2 different sets of data, and happen to share a part of their UI.

salvamomo (commented):
The count check on the $options array to select a value by default looks fine, thanks! 👍 

If we end up going with an abstract class , I think the `getFormId` method should be removed from there and placed only in the child classes.

On the rest, I'm going to let someone else make the decision because I don't think these classes call for an abstract class. To me, that pattern makes sense when there's some sort of flow or logic when using a service, which follows a similar structure in which several methods need to be called, but most of those methods change depending on the class being called. This is not the case here, as there isn't any `abstract` method in the class being introduced. 
 
For the `buildForm` methods, as I think they should be completely split (thus avoid the parent::buildForm()) even if they happen to have 2 fields that are same. I see the intent: "If in the future I need a change that applies to both forms, I only do it in the parent class and that's done".  However to me using this pattern in forms makes it easier to make mistakes that go unnoticed: "If in the future I need a change that applies to one form, I can apply it by mistake in the parent form. Tests will still probably pass, the UI of the form I'm testing will look ok, but it'll be harder to spot in the diff that the change was done in the wrong class and that is also appearing in others" .

Extracting out `loadVideoTypeOptions` makes total sense 💯 . I think it's better placed in a separate service `MpxFormUtils` or similar, than in the abstract class, for the reason mentioned, or maybe on a Trait, but otherwise, great call on trying to share that bit :). 

Not approving or requesting changes, as some more people work on this module, and this is obviously my take on it, but they might have different views 🙃. Just wanted to leave my thoughts, since to me reducing these lines of code is not worth the additional inheritance layer a dev needs to know about when working on the classes.

juampynr:
I am with @salvamomo. By using an abstract class for the form the code is shorter but also harder to understand and maintain.

edutrul:
Makes sense! thanks @salvamomo and @juampynr 
Better use traits for this case. The trait created is for creating base form and for a helper function to list available mpx accounts.

salvamomo (commented):
Could we still leave the build form logic in each of their forms. I really don't see any problem with repeating a couple of form API arrays, and it's just simpler for any dev to do work on the forms. But if @juampynr thinks this is good enough as it is, fine by me! 👍

src/Form/ImportUpdateVideoItemTrait.php
```diff
@@ -0,0 +1,65 @@
...
+      '#description' => $this->t('Choose the video type to import the video into.'),
+      '#options' => $video_opts,
+      '#default_value' => count($video_opts) === 1 ? array_keys($video_opts) : [],
+      '#required' => TRUE,
+    ];
+    $form['actions']['#type'] = 'actions';
+    $form['actions']['submit'] = [
+      '#type' => 'submit',
+      '#value' => $this->t('Import video'),
+      '#button_type' => 'primary',
```
deviantintegral, line 35: Interesting we have to define this.

src/Form/ImportUpdateVideoItemTrait.php
```diff
@@ -0,0 +1,65 @@
...
+ * @package Drupal\media_mpx\Form
+ */
+trait ImportUpdateVideoItemTrait {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildBaseForm(array $form) {
+    if (!$video_opts = $this->loadVideoTypeOptions()) {
+      $this->messenger()->addError($this->t('There has been an unexpected problem loading the form. Reload the page.'));
```
deviantintegral, line 19: Weird, in that I don't see how reloading the form would fix anything. It's fine as I see this is in existing, code, just 🤔 .

src/Form/ImportUpdateVideoItemTrait.php
```diff
@@ -0,0 +1,65 @@
...
+    $video_opts = [];
+
+    try {
+      $video_types = $this->mpxTypeRepository->findAllTypes();
+
+      foreach ($video_types as $type) {
+        $video_opts[$type->id()] = $type->label();
+      }
+    }
+    catch (\Exception $e) {
```
deviantintegral, line 58: Yeah this seems like a code smell.

deviantintegral (approved):
A few notes, but really just thoughts that may be nothing or could be done in followups. Nice!

(also, what a hilariously long branch name 😂 )

PR #155, opened 2019-10-23
Work around CircleCI permission issues with the artifacts folder

m4olivei:
Workaround to make sure the artifacts folder in CirclCI runs has write permissions.


deviantintegral (approved):
I like that the sniffer detects misnamed class files!

PR #157, opened 2019-11-25
Issue #156 Force a fresh request to mpx when performing an update via the UI 

### LINKED ISSUE #156: Caching of mpx responses causes unexpected behavior for the 'Update mpx data' and 'Import mpx video by <id>' forms
From the README, there is caching at play for any for any request to mpx for a single data object. As such, when you use the 'Update mpx data' button, ot the 'Import mpx video by <id>' forms, it will work the way you expect the first time, but will also warm that cache. On subsequent uses of those UI elements, the cached data will be returned until the cache item expires.

We need to force a bypass of the cache for these UI elements so that the absolute most hot and fresh baked data comes back from mpx.

:bread:

m4olivei:
Resolves #156.

### DIFF
--- a/src/Service/UpdateVideoItem/UpdateVideoItem.php
+++ b/src/Service/UpdateVideoItem/UpdateVideoItem.php
@@ -69,7 +69,7 @@ class UpdateVideoItem {
 
     $media_source = $this->importer->loadMediaSource($media_type);
     $factory = $this->objectFactoryCreator->fromMediaSource($media_source);
-    $results = $factory->loadByNumericId($mpx_id);
+    $results = $factory->loadByNumericId($mpx_id, FALSE, ['headers' => ['Cache-Control' => 'no-cache']]);
     $mpx_media = $results->wait();
     $saved = $this->importer->importItem($mpx_media, $media_type);
 


src/Service/UpdateVideoItem/UpdateVideoItem.php
```diff
@@ -69,7 +69,7 @@ public function execute(UpdateVideoItemRequest $request): UpdateVideoItemRespons
 
     $media_source = $this->importer->loadMediaSource($media_type);
     $factory = $this->objectFactoryCreator->fromMediaSource($media_source);
-    $results = $factory->loadByNumericId($mpx_id);
+    $results = $factory->loadByNumericId($mpx_id, FALSE, ['headers' => ['Cache-Control' => 'no-cache']]);
```
mattwith, line 72: 👍

mattwith:
This looks great! I've verified this resolves the caching we've seen on some downstream sites.

src/Service/UpdateVideoItem/UpdateVideoItem.php
salvamomo, line 72: Great catch @m4olivei !

PR #159, opened 2019-12-03
Issue #158 Add a console link field formatter

### LINKED ISSUE #158: Add a field formatter to link to the mpx console
It would be dandy if there were a field formatter, which when applied to a string field on an mpx entiity, it would link to the same media in the mpx console, eg.

https://console.theplatform.com/<account_id>/media/<guid>/metadata#information

m4olivei:
Resolves #158

### DIFF
--- a/src/Plugin/Field/FieldFormatter/ConsoleLinkFormatter.php
+++ b/src/Plugin/Field/FieldFormatter/ConsoleLinkFormatter.php
@@ -16,7 +16,7 @@ use Drupal\media_mpx\Plugin\media\Source\Media as MpxMediaSource;
  * @FieldFormatter(
  *   id = "media_mpx_console_link",
  *   label = @Translation("mpx Console Link"),
- *   description = @Translation("Suitable for use on fields storing the mpx guid value, the mpx console link formatter will create a link to the media in the mpx console.")
+ *   description = @Translation("Suitable for use on fields storing the mpx guid value, the mpx console link formatter will create a link to the media in the mpx console."),
  *   field_types = {
  *     "string"
  *   }
--- a/src/Plugin/Field/FieldFormatter/ConsoleLinkFormatter.php
+++ b/src/Plugin/Field/FieldFormatter/ConsoleLinkFormatter.php
@@ -2,8 +2,6 @@
 
 namespace Drupal\media_mpx\Plugin\Field\FieldFormatter;
 
-use Drupal\Core\Annotation\Translation;
-use Drupal\Core\Field\Annotation\FieldFormatter;
 use Drupal\Core\Field\FieldItemInterface;
 use Drupal\Core\Field\FieldItemListInterface;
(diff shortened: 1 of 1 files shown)


src/Plugin/Field/FieldFormatter/ConsoleLinkFormatter.php
```diff
@@ -0,0 +1,99 @@
...
+    $path_parts = explode('/', $account->getMpxId()->getPath());
+    $account_id = end($path_parts);
+    $options = [
+      'attributes' => [
+        'target' => '_blank',
+        'rel' => 'nofollow',
+      ],
+    ];
+    return Url::fromUri(
+      sprintf('https://console.theplatform.com/%s/media/%s/metadata#information', $account_id, $item->getString()),
```
salvamomo, line 94: Being super picky, my only suggestion here would be to perhaps define this string as a const, or something more visible instead of left here, but this is a **very** specific formatter, so feel free to just ignore!

salvamomo (approved):
LGTM!

deviantintegral:
Does this make more sense in the mpx-php library? I could see there being a `ConsoleLinkInterface` with a `getConsoleLink`, that all mpx objects could implement including playlists, feeds, etc.

However, I think we could do that later without breaking this class - just the internal code to actually generate the link would change, so I'm 👍 on this.

m4olivei:
> Does this make more sense in the mpx-php library? I could see there being a ConsoleLinkInterface with a getConsoleLink, that all mpx objects could implement including playlists, feeds, etc.

Oh, nice idea. I like that. I'll file an issue for a rainy day on that.

m4olivei:
Filed upstream issue: https://github.com/Lullabot/mpx-php/issues/164

src/Plugin/Field/FieldFormatter/ConsoleLinkFormatter.php
```diff
@@ -0,0 +1,99 @@
...
+   *
+   * @return bool
+   *   TRUE if this field formatter applies, otherwise FALSE.
+   */
+  protected function appliesTo(FieldItemListInterface $items) {
+    if (!($entity = $items->getEntity()) || !$entity instanceof Media) {
+      return FALSE;
+    }
+
+    $source_plugin = $entity->getSource();
```
edutrul, line 65: Nitpick: It'd be great to add @var 
```
/* @var \Drupal\media_mpx\Plugin\media\Source\Media $source_plugin **/
```

m4olivei: It's not really needed. Since the next line is a conditional checking the class, PHPStorm will infer the type beyond that point in the code and still give you autocomplete. I'm going to skip this if it's alright, since its'a bit of a pain to update all three relevant PR's in play.

edutrul: sounds good

PR #164, opened 2021-06-22
Drupal 9 Compatibility

m4olivei:
Update the module for Drupal 9 compatibility.

Start with adding a job to run `drupal-check`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,6 +127,10 @@ functional_js_tests: &functional_js_tests
         command: |
           ./test-functional-js.sh $CIRCLE_PROJECT_REPONAME
 
+    - run:
+        command: |
+          [ -d /var/www/html/sites/simpletest/browser_output ] && echo 'Copying functional test browser output to artifacts directory...' && cp -R /var/www/html/sites/simpletest/browser_output /var/www/html/artifacts
+
     - store_test_results:
         path: /var/www/html/artifacts/phpunit
     - store_artifacts:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,7 +85,6 @@ unit_kernel_tests: &unit_kernel_tests
     - run:
         working_directory: /var/www/html
         command: |
-          chown -R www-data artifacts
           ./test.sh $CIRCLE_PROJECT_REPONAME
 
     - store_test_results:
@@ -106,7 +105,6 @@ functional_tests: &functional_tests
     - run:
         working_directory: /var/www/html
         command: |
-          chown -R www-data artifacts
           ./test-functional.sh $CIRCLE_PROJECT_REPONAME
 
     - store_test_results:
@@ -127,7 +125,6 @@ functional_js_tests: &functional_js_tests
     - run:
         working_directory: /var/www/html
         command: |
-          chown -R www-data artifacts
           ./test-functional-js.sh $CIRCLE_PROJECT_REPONAME
 
     - store_test_results:
@@ -183,7 +180,6 @@ code_coverage: &code_coverage
     - run:
         working_directory: /var/www/html
         command: |
-          chown -R www-data artifacts
           ./code-coverage-stats.sh $CIRCLE_PROJECT_REPONAME
     - store_artifacts:
         path: /var/www/html/artifacts
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -199,7 +199,7 @@ deprecated_code_check: &deprecated_code_check
     - run:
         working_directory: /var/www/html
         command: |
-          vendor/bin/drupal-check -d -e modules/media_mpx
+          vendor/bin/drupal-check -d modules/media_mpx
 
 # Declare all of the jobs we should run.
 version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -188,6 +188,19 @@ code_coverage: &code_coverage
     - store_artifacts:
         path: /var/www/html/artifacts
 
+deprecated_code_check: &deprecated_code_check
+  <<: *defaults
+  steps:
+    - attach_workspace:
+        at: /var/www/html
+
+    - checkout
+
+    - run:
+        working_directory: /var/www/html
+        command: |
+          vendor/bin/drupal-check -d -e modules/media_mpx
+
 # Declare all of the jobs we should run.
 version: 2
 jobs:
@@ -205,6 +218,8 @@ jobs:
      <<: *code_sniffer
   run-code-coverage:
      <<: *code_coverage
+  run-deprecated-code-check:
+     <<: *deprecated_code_check
 
 workflows:
   version: 2
@@ -232,6 +247,9 @@ workflows:
           requires:
             - update-dependencies
             - run-unit-kernel-tests
+      - run-deprecated-code-check:
+          requires:
+            - update-dependencies
 
   weekly:
     jobs:
@@ -255,6 +273,9 @@ workflows:
           requires:
             - update-dependencies
             - run-unit-kernel-tests
+      - run-deprecated-code-check:
+          requires:
+            - update-dependencies
 
     triggers:
       - schedule:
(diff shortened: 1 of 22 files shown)


m4olivei:
Yay! Failed! Finally got the job right :).

m4olivei:
Getting so close. Here is the error we're hitting in the tests:

```
Drupal\Tests\media_mpx\Kernel\Access\MediaAvailableAccessTest::testAccessChecksForMediaItemWithTimestampFields
Trying to access array offset on value of type null

/home/vagrant/code/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/DocParser.php:995
/home/vagrant/code/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/DocParser.php:687
/home/vagrant/code/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/DocParser.php:663
/home/vagrant/code/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/DocParser.php:354
/home/vagrant/code/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/AnnotationReader.php:401
/home/vagrant/code/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/AnnotationReader.php:334
/home/vagrant/code/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/AnnotationReader.php:214
/home/vagrant/code/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/AnnotationReader.php:226
/home/vagrant/code/vendor/lullabot/mpx-php/src/DataService/DataServiceDiscovery.php:100
/home/vagrant/code/vendor/lullabot/mpx-php/src/DataService/DataServiceDiscovery.php:81
/home/vagrant/code/vendor/lullabot/mpx-php/src/DataService/DataServiceManager.php:58
/home/vagrant/code/web/modules/contrib/media_mpx/src/Plugin/media/Source/Media.php:78
/home/vagrant/code/web/modules/contrib/media_mpx/src/Plugin/media/Source/MediaSourceBase.php:145
/home/vagrant/code/web/core/lib/Drupal/Core/Plugin/Factory/ContainerFactory.php:21
/home/vagrant/code/web/core/lib/Drupal/Component/Plugin/PluginManagerBase.php:83
/home/vagrant/code/web/modules/contrib/media_mpx/tests/src/Kernel/MediaMpxTestBase.php:87
/home/vagrant/code/web/modules/contrib/media_mpx/tests/src/Kernel/Access/MediaAvailableAccessTest.php:44
```
There are many similar errors. They all stem from an issue with `doctrine/annotations` and PHP 7.4: https://github.com/doctrine/annotations/issues/273. We need to update to a more recent version of `doctrine/annotations`, however, Drupal 8.9.16, specifically `drupal/core-recommended:8.9.16` pins `doctrine/annotations` to `1.4.0` (https://packagist.org/packages/drupal/core-recommended#8.9.16). Not sure what to do about that just yet. For some sites (including the one I'm working on) this is a non-issue b/c `drupal/core` is required directly which doesn't pin the version and lets you get something more recent.

m4olivei:
Note: Details regarding recent commit to require `kevinrob/guzzle-cache-middleware:~3.3.1` are here [kevinrob/guzzle-cache-middleware](https://github.com/Kevinrob/guzzle-cache-middleware/pull/143#issuecomment-877700383)

.circleci/config.yml
```diff
@@ -16,7 +16,7 @@
 defaults: &defaults
   docker:
     # specify the version you desire here (avoid latest except for testing)
-    - image: quay.io/deviantintegral/drupal_tests:0.5.0-drupal87
+    - image: quay.io/m4olivei/drupal-8.9-test:latest
```
m4olivei: Will change this once `drupal_tests` is all approved, merged and tagged.

.circleci/config.yml
```diff
@@ -183,11 +182,24 @@ code_coverage: &code_coverage
     - run:
         working_directory: /var/www/html
         command: |
-          chown -R www-data artifacts
           ./code-coverage-stats.sh $CIRCLE_PROJECT_REPONAME
     - store_artifacts:
         path: /var/www/html/artifacts
 
+deprecated_code_check: &deprecated_code_check
```
deviantintegral, line 189: ❤️

composer.json
```diff
@@ -15,22 +15,22 @@
   },
   "require": {
     "drupal/guzzle_cache": "^2.0",
+    "kevinrob/guzzle-cache-middleware": "~3.3.1",
```
deviantintegral, line 18: Interesting, isn't this required due to guzzle_cache above? Also, FYI there's an API-compatible 3.0 release of guzzle_cache, so you could `||` that in the require line.

src/CustomFieldDiscovery.php
```diff
@@ -62,6 +62,7 @@ public function getCustomFields(): array {
     // Clear the annotation loaders of any previous annotation classes.
     AnnotationRegistry::reset();
     // Register the namespaces of classes that can be used for annotations.
+    // @phpstan-ignore-next-line
```
deviantintegral, line 68: Perhaps document here and below _why_ the lines are being ignored?

tests/src/FunctionalJavascript/PlayerFormatterTest.php
```diff
@@ -95,16 +100,12 @@ public function testViewPlayerFormatter() {
     // Add the new field to the default form and view displays for this
     // media type.
     if ($source_field->isDisplayConfigurable('form')) {
-      // @todo Replace entity_get_form_display() when #2367933 is done.
```
deviantintegral, line 98: 👏

deviantintegral (approved):
Wow, is this working and done? I am amazed with how little there actually is here to update. A great example of a fairly complex module that didn't take _that_ much work to update.

composer.json
m4olivei: > Interesting, isn't this required due to guzzle_cache above?

It is, although when I was working on this, it would pick up a version of `kevinrob/guzzle-cache-middleware` that implicitlyl depended on a recent version of `guzzlehttp/psr7` without expressing that properly in it's requirements, which led to situations, like here, where it was using recent API's in `guzzlehttp/psr7` that weren't actually present. More details here: https://github.com/Kevinrob/guzzle-cache-middleware/pull/143#issuecomment-877700383

The good news is that the author made a new release with the right requirements, so I'm going to try to remove this and see if we're all good.

.circleci/config.yml
m4olivei: Updated!

composer.json
m4olivei: Nope. Had to stick with `~3.3.1`. What it comes down to is that `drupal/core-recommended` (which the test image is based on) pins `guzzlehttp/psr7` at 1.6.1, which 3.3.1 of `kevinrob/guzzle-cache-middleware` is the last version to be compatible with that.

deviantintegral: I think if you do `^3.3`, it will allow you to upgrade once the psr7 constraint is lifted. `~3.3.1` _adds_ a constraint preventing upgrading to 3.4 or higher.

I think!

m4olivei: Yep, `~3.3.1` means it can't go >= `3.4`. That's intentional ATM as with something like `^3.3` composer wants to go to `3.4`, but `3.4` has a hidden dependency on a version of `guzzlehttp/psr7` that is too recent for `drupal/core-recommended` to allow. `3.4.1` has the hidden dependency declared explicitly, but that means composer won't take it b/c it conflicts with `drupal/core-recommended`. So we're stuck sticking to the `3.3.x` line until a major update unfortunately that shirks D8.

deviantintegral: > a hidden dependency on a version of `guzzlehttp/psr7`

_ohhhh_ ok, that makes sense. At a moment like this I really wish you could have comments in JSON, but at least we'll have this thread for future us to remember!

m4olivei: Totally! I definately googled "comments in composer.json" along the way here :laughing:

